PR #30, opened 2018-09-19
Create e2e setup for specific operator

MarcelMue:
towards https://github.com/giantswarm/giantswarm/issues/3313

The operator gets installed and removed successfully :tada:


integration/env/common.go
```diff
@@ -44,3 +53,7 @@ func CircleSHA() string {
 func KeepResources() string {
 	return keepResources
 }
+
+func RegistryPullSecret() string {
+	return registryPullSecret
+}
```
xh3b4sd: DO we have the explicit pull secret magic in the other operators as well? Why do we need this here now? Wasn't this magically done somewhere else? How do you want to roll with this?

integration/test/basic/main_test.go
```diff
@@ -84,17 +85,10 @@ func init() {
...
-			l.LogCtx(ctx, "level", "error", "message", "e2e test setup failed", "stack", fmt.Sprintf("%#v\n", err))
-		}
-
-		os.Exit(v)
+	v, err := setup.WrapTestMain(h, helmClient, l, m)
+	if err != nil {
+		l.LogCtx(ctx, "level", "error", "message", "e2e test failed", "stack", fmt.Sprintf("%#v\n", err))
 	}
+
+	os.Exit(v)
```
xh3b4sd: This integer dispatching and exit management is awful as hell. I cannot help but it just feels painful. How can we improve this? What is the actual problem we try to address with it?

integration/env/common.go
MarcelMue: Most operators have it, you didn't see it before because it was in some hidden env variable before. See `aws-operator` which does it the same way I did:
https://github.com/giantswarm/e2etemplates/blob/master/pkg/chartvalues/aws_operator.go#L17

And `azure-operator` which has it still in some hidden env variable:
https://github.com/giantswarm/azure-operator/blob/master/integration/template/azure_operator_charts_values.go#L46

So I would say that it's progress and we atleast have some vision on it.

integration/test/basic/main_test.go
MarcelMue: The issue we had here was that we only want to really run a test if the setup was successful, but still want to do the teardown even if it wasn't successful. And then we want to have errors from all stages but preserve the exit code if something along the way went wrong.

It can definitely be improved but I'd like to do that in a separate PR.

PR #33, opened 2018-09-21
adapts desired state to CR changes

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/3313.


Gopkg.toml
```diff
@@ -24,13 +24,20 @@
 #   go-tests = true
 #   unused-packages = true
 
+
+
 required = [
-  "github.com/docker/distribution",
+  "github.com/docker/distribution/reference",
```
xh3b4sd: Don't know. This was funky now.

service/controller/v1/key/key.go
```diff
@@ -1,18 +1,16 @@
...
 
 const (
-	LabelApp          = "app"
-	LabelManagedBy    = "giantswarm.io/managed-by"
-	LabelOrganization = "giantswarm.io/organization"
-	LabelServiceType  = "giantswarm.io/service-type"
+	LabelApp            = "app"
+	LabelManagedBy      = "giantswarm.io/managed-by"
+	LabelOrganization   = "giantswarm.io/organization"
+	LabelReleaseVersion = "release.giantswarm.io/version"
```
xh3b4sd: We have to add the release version as label to the `ChartConfig` CRs in order to fetch the current state easily.

service/controller/v1/key/key.go
```diff
@@ -21,24 +19,8 @@ const (
 	ServiceTypeManaged = "managed"
 )
 
-func OperatorChannelName(customResource v1alpha1.Release) string {
```
xh3b4sd: Lots of information are based on the "authorities" now and I am thinking that we should move more key functions to `apiextensions` in general.

service/controller/v1/resource/chartconfig/current.go
```diff
@@ -30,16 +29,16 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
+	//	if err != nil {
+	//		return nil, microerror.Mask(err)
+	//	}
 
 	var chartConfigCR *v1alpha1.ChartConfig
 	{
 		r.logger.LogCtx(ctx, "level", "debug", "message", "finding current state")
 
-		m, err := r.g8sClient.CoreV1alpha1().ChartConfigs(r.namespace).Get(key.OperatorChartName(customResource), metav1.GetOptions{})
+		m, err := r.g8sClient.CoreV1alpha1().ChartConfigs(r.namespace).Get("", metav1.GetOptions{})
```
xh3b4sd: Just making it compile now. Fixing in the next PRs.

service/controller/v1/resource/chartconfig/desired.go
```diff
@@ -41,30 +50,31 @@ func (r *Resource) newChartCR(ctx context.Context, customResource v1alpha1.Relea
...
-				Name:      key.OperatorChartName(customResource),
+				Name:      fmt.Sprintf("%s-chart", authority.Name),
 				Namespace: metav1.NamespaceSystem,
 				Secret: v1alpha1.ChartConfigSpecSecret{
 					Name:            c.Secret.Name,
 					Namespace:       c.Secret.Namespace,
 					ResourceVersion: c.Secret.ResourceVersion,
 				},
-				Release: key.ReleaseName(customResource),
+				Release: fmt.Sprintf("%s-%s", authority.Name, strings.Replace(authority.Version, ".", "-", -1)),
```
xh3b4sd: I figured that the `Channel` and the `Release` are actually the same in our case here. Can anybody confirm?

MarcelMue: Yes this should be the case because we only expect to run one release per channel.

service/controller/v1/key/key.go
MarcelMue: Moving most of these functions to apiextensions would make sense to me.

service/controller/v1/resource/chartconfig/desired.go
MarcelMue: This fmt stuff should also be inside the key package IMO.

xh3b4sd: I will go with this and provide PRs to use key functions from `apiextensions`.

service/controller/v1/key/key.go
rossf7: Cool to have this label. I can see it being useful for lots of things.

service/controller/v1/resource/chartconfig/desired.go
rossf7: Not into the details but that makes sense to me. 

Stating the obvious but release is used as the Helm release name. One gotcha we hit was you can't have multiple releases with the same name. Even if they are installed in different namespaces.

xh3b4sd: See https://github.com/giantswarm/apiextensions/pull/143.

PR #43, opened 2019-01-17
app resource

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/5112

Layout basics for the new app resource in charge of creating app CRs.

### DIFF
--- a/service/controller/v1/resource/app/create.go
+++ b/service/controller/v1/resource/app/create.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"context"
+	"fmt"
 
 	applicationv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/microerror"
@@ -15,7 +16,7 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 	}
 
 	if len(appCRs) != 0 {
-		r.logger.LogCtx(ctx, "level", "debug", "message", "creating App CRs in the Kubernetes API")
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating %d App CRs in the Kubernetes API", len(appCRs)))
 
 		for _, c := range appCRs {
 			_, err = r.g8sClient.ApplicationV1alpha1().Apps(r.namespace).Create(c)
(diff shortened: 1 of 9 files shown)


service/controller/v1/resource/app/desired.go
```diff
@@ -0,0 +1,80 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: authority.HelmReleaseName(),
+			Labels: map[string]string{
+				key.LabelApp:            authority.Name,
+				key.LabelManagedBy:      key.ProjectName,
+				key.LabelOrganization:   key.OrganizationName,
+				key.LabelReleaseVersion: key.ReleaseVersion(releaseCR),
+				key.LabelServiceType:    key.ServiceTypeManaged,
+			},
```
rossf7: We'll need to do something for internal versioning for app CRs but its not agreed yet. At the moment chart-operator uses an annotation `giantswarm.io/version-bundle` but its likely that will change.

@TheoBrigitte I think its best we leave it like this for now. I'll let you know once this is agreed in the App Catalog spec. Sorry about the hassle.

TheoBrigitte: Nothing was agreed at the time I wrote this code.

But now we have a new `OPERATOR.giantswarm.io/version` label in fmt https://github.com/giantswarm/fmt/blob/master/kubernetes/annotations_and_labels.md#common-labels

Here we could have something like: `app-operator.giantswarm.io/version=1.0.0`

rossf7: @TheoBrigitte Yes we discussed this in Batman planning and we'll use the label `app-operator.giantswarm.io/version=1.0.0`.

I still need to get the spec PR merged. But we might as well get it sorted now rather than delay.

service/controller/v1/resource/app/create.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
+	appCRs, err := toAppCRs(createChange)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if len(appCRs) != 0 {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "creating App CRs in the Kubernetes API")
```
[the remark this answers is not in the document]
TheoBrigitte: added number of app CRs being created

service/controller/v1/resource/app/delete.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange interface{}) error {
+	appCRs, err := toAppCRs(deleteChange)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if len(appCRs) != 0 {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "deleting App CRs in the Kubernetes API")
```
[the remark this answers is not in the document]
TheoBrigitte: added number of app CRs being updated

service/controller/v1/resource/app/desired.go
TheoBrigitte: We also have an ongoing discussion in Magic in order to know where we get the app-operator version from, and how its managed through updated.

So I added the `app-operator.giantswarm.io/version` label with no value for now. We will sort the value in a different PR.

PR #45, opened 2019-02-06
Align with new ReleaseCycle and App CR changes

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/5277

* update k8s dependencies to `kubernetes-1.13.2`
* update vendor
* fix metricsresource.WrapConfig

* watch ReleaseCycle CR
* align App CR changes
* drop version bundle

The goal here is to get rid of the old Release CR and introducing the new ReleaseCycle CR.


service/controller/v1/resource/app/desired.go
```diff
@@ -4,15 +4,20 @@ import (
...
 )
 
 func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error) {
-	releaseCR, err := key.ToCustomResource(obj)
+	releaseCycleCR, err := key.ToReleaseCycleCR(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	releaseCR, err := r.g8sClient.ReleaseV1alpha1().Releases(releaseCycleCR.GetNamespace()).Get(releaseCycleCR.Spec.Release.Name, metav1.GetOptions{})
```
TheoBrigitte: This is far from the logic we want to have here. I will update it in later PR.

service/controller/v1/resource/chartconfig/current.go
```diff
@@ -30,7 +30,12 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 		return nil, microerror.Mask(err)
 	}
 
-	customResource, err := key.ToCustomResource(obj)
+	releaseCycleCR, err := key.ToReleaseCycleCR(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	releaseCR, err := r.g8sClient.ReleaseV1alpha1().Releases(releaseCycleCR.GetNamespace()).Get(releaseCycleCR.Spec.Release.Name, metav1.GetOptions{})
```
TheoBrigitte: This is far from the logic we want to have here. I will update it in later PR.

service/controller/v1/resource/chartconfig/desired.go
```diff
@@ -12,17 +13,22 @@ import (
...
+	releaseCycleCR, err := key.ToReleaseCycleCR(obj)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
 
-	var desiredChartConfigCRs []*v1alpha1.ChartConfig
+	releaseCR, err := r.g8sClient.ReleaseV1alpha1().Releases(releaseCycleCR.GetNamespace()).Get(releaseCycleCR.Spec.Release.Name, metav1.GetOptions{})
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
```
TheoBrigitte: This is far from the logic we want to have here. I will update it in later PR.

service/controller/v1/resource_set.go
```diff
@@ -126,9 +125,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 	}
 
 	{
-		c := metricsresource.WrapConfig{
-			Name: config.ProjectName,
```
TheoBrigitte: Updated according to : https://github.com/giantswarm/operatorkit/pull/276

service/controller/v1/resource_set.go
```diff
@@ -144,24 +141,11 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
...
-			return false
-		}
-
-		if key.VersionBundleVersion(customResource) == VersionBundle().Version {
-			return true
-		}
-
-		return false
-	}
-
```
TheoBrigitte: We are moving from version bundle to label selector. Dropping this for now and will add label selector in later PR.

service/controller/v1/key/key.go
```diff
@@ -20,20 +20,16 @@ const (
...
+		return releasev1.ReleaseCycle{}, microerror.Maskf(wrongTypeError, "expected '%T', got '%T'", &releasev1.ReleaseCycle{}, v)
 	}
 	customResource := *customResourcePointer
 
 	return customResource, nil
 }
-
-func VersionBundleVersion(customResource v1alpha1.Release) string {
-	return customResource.Spec.VersionBundle.Version
-}
```
TheoBrigitte: We are moving from version bundle to label selector.

service/controller/v1/resource/chartconfig/desired.go
```diff
@@ -3,7 +3,8 @@ package chartconfig
 import (
 	"context"
 
-	"github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
+	corev1 "github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
+	releasev1 "github.com/giantswarm/apiextensions/pkg/apis/release/v1alpha1"
```
xh3b4sd: The alias does not express the truth. When we want to prefix, it should be `releasev1alpha1`. The important path is not `.../release/v1`.

service/controller/v1/resource_set.go
xh3b4sd: How does the deployment work though? I feel like dropping this here should be the last step to activate the new behaviour.

TheoBrigitte: This operator is not deployed yet.

service/controller/v1/resource/chartconfig/desired.go
TheoBrigitte: Good point.

Renamed all:
applicationv1 > applicationv1alpha1
corev1 > corev1alpha1
releasev1 > releasev1alpha1

PR #46, opened 2019-02-07
resource/chartconfig: delete

kopiczko:
This operator is not going to create any chartconfigs.


kopiczko:
They don't use the most recent advancements. So they fail because of timing issues.

xh3b4sd:
Alright but let's get this straight then. When the operator is rewritten we maybe can just drop the failing tests for now and start from scratch ones we have anything to verify.

kopiczko:
> Alright but let's get this straight then. When the operator is rewritten we maybe can just drop the failing tests for now and start from scratch ones we have anything to verify.

Those test are doing nothing AFAICT. We'll override them anyway.

PR #47, opened 2019-02-11
Install Release CRD

TheoBrigitte:

Ensure Release CRD is created when the operator boots up.


service/controller/release.go
```diff
@@ -51,7 +51,7 @@ func NewRelease(config ReleaseConfig) (*Release, error) {
 	{
 		c := informer.Config{
 			Logger:  config.Logger,
-			Watcher: config.G8sClient.Release().ReleaseCycles(""),
+			Watcher: config.G8sClient.ReleaseV1alpha1().ReleaseCycles(""),
```
MarcelMue: This should already make sure that the CRD exists or am I not seeing something? 
`Operatorkit` makes sure that the crd exists for the CR we are watching.

TheoBrigitte: This is the ReleaseCycle CRD, but this PR is about Release CRD.

service/service.go
```diff
@@ -137,6 +162,20 @@ func New(config Config) (*Service, error) {
 // Boot starts top level service implementation.
 func (s *Service) Boot() {
 	s.bootOnce.Do(func() {
-		go s.releaseController.Boot(context.Background())
+		ctx := context.Background()
+		backOffFactory := func() backoff.Interface { return backoff.NewMaxRetries(7, 1*time.Second) }
```
kopiczko: Why this is a factory and not just a backoff value?

service/service.go
```diff
@@ -137,6 +162,20 @@ func New(config Config) (*Service, error) {
...
+
+		err := s.crdClient.EnsureCreated(ctx, s.crd, backOffFactory())
+		if err != nil {
+			s.logger.LogCtx(ctx, "level", "error", "message", "stop service boot retries due to too many errors", "stack", fmt.Sprintf("%#v", err))
+			os.Exit(1)
+		}
+
+		s.logger.LogCtx(ctx, "level", "debug", "message", "ensured custom resource definition exists")
+
+		go s.releaseController.Boot(ctx)
```
kopiczko: I don't remember exact reason for doing that but and I think we shouldn't do that in general. Let's try to skip this `go ` this time and see what happens.

@xh3b4sd maybe you remember what was wrong with calling it synchronously?

service/service.go
```diff
@@ -137,6 +162,20 @@ func New(config Config) (*Service, error) {
...
 func (s *Service) Boot() {
 	s.bootOnce.Do(func() {
-		go s.releaseController.Boot(context.Background())
+		ctx := context.Background()
+		backOffFactory := func() backoff.Interface { return backoff.NewMaxRetries(7, 1*time.Second) }
+
+		// Install Release CRD.
+		s.logger.LogCtx(ctx, "level", "debug", "message", "ensuring custom resource definition exists")
+
+		err := s.crdClient.EnsureCreated(ctx, s.crd, backOffFactory())
```
xh3b4sd: That is not right. Why are we doing this here? The controller does it already. We should not do it manually.

service/service.go
xh3b4sd: Well, the controller watches events and `Boot` can always block until it is done and you may don't want that. Everything in `Boot` should be executed concurrently IMO.

service/service.go
kopiczko: This is right. The controller creates ReleaseCycle CRD. This here is Release CRD.

service/service.go
kopiczko: Oh yeah. Because `Boot` is blocking. We need to get that stuff straight at some point.

service/service.go
TheoBrigitte: Here we are installing the `ReleaseCRD` while the controller installs the `ReleaseCycleCRD`.

I did not fond another way to have the operator install a second CRD.

xh3b4sd: The way to install a CRD in our case is to boot a controller. Why aren't we doing simply this?

TheoBrigitte: Because the operator doesn't actually need this second controller, so it feels weird to boot a whole controller that we are not going to use.

service/service.go
TheoBrigitte: Good point, I did a lazy copy paste here.

Simplified it. PTAL

service/service.go
TheoBrigitte: Does this mean that I should also wrap the crd creation in a separate go routine here ?

kopiczko: No I got confused. Forget about it.

service/service.go
```diff
@@ -137,6 +162,20 @@ func New(config Config) (*Service, error) {
...
 func (s *Service) Boot() {
 	s.bootOnce.Do(func() {
-		go s.releaseController.Boot(context.Background())
+		ctx := context.Background()
+		backOff := backoff.NewMaxRetries(7, 1*time.Second)
+
+		// Install Release CRD.
+		s.logger.LogCtx(ctx, "level", "debug", "message", "ensuring release custom resource definition exists")
+
+		err := s.crdClient.EnsureCreated(ctx, s.crd, backOff())
```
tomahawk28: If we are going to install `Release` CRD in below link, https://github.com/giantswarm/operatorkit/blob/a1aa0b72cce3bb752ca8e5097dc84f97cf4cbbf7/controller/controller.go#L367
(operator-kit, controller module) 
Why we need to do that again in here?

TheoBrigitte: This is about installing `Release` CRD and not `ReleaseCycle` CRD

service/service.go
xh3b4sd: When the operator does not watch the CR it should not manage the CRD. Simple as that. When the Release CRD is a common resource we expect to be there it might should be bootstrapped by the control plane.

TheoBrigitte: This operator do reconciles `Release` CR and `ReleaseCycle` CR.

It watches `ReleaseCycle` CR and ensure corresponding `Release` CR are installed.
Then reconciles `Release` CR to have their components installed.

So release-operator is responsible for `Release` CRD, in the sense that is do the reconcile `Release` CR.
It would feel strange to have "something" else installing the `Release` CRD.

xh3b4sd: Sounds like you need a controller so I am not sure what we are talking about here.

xh3b4sd: For the record. We had a quick hang with Theo. Creating exceptions here smells. The idea now was to have a controller for the `Release` CRs and have a `releasestatus` resource reconciled in order to manage the CR status. The controller ensures the `Release` CRD. That is straight forward, clean and simple. `Release` and `ReleaseCycle` need to cross reference their 1 to 1 relationship though, which should not be hard. In case you would want to separate the controllers later on it would be as easy as moving files around. No need to deal with resource and test refactorings.

kopiczko: I don't see a point in running a watcher running for Release CRs. It is an overkill for just creating a CRD.

xh3b4sd: It is for updating the CR status separately. I think this makes sense and is a clean approach without ugly hacks and exceptions.

kopiczko: We still need to watch ReleaseCycle CRs to update status. There is a reference from ReleaseCycle CR to Release CR but no other way around.

TheoBrigitte:
Mr @kopiczko and me talked, we agreed on having a second controller for `Release` CR. Therefore the `Release` CRD will be created by this controller. Let the magic happen.

xh3b4sd:
magic

PR #48, opened 2019-02-12
app: consider namespace when comparing App CRs

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/5282.


service/controller/v1/resource/app/resource.go
```diff
@@ -66,23 +66,14 @@ func (r *Resource) Name() string {
...
-	_, err := getAppCRByName(list, item.Name)
-	if err != nil {
-		return false
-	}
-
-	return true
-}
-
-func getAppCRByName(list []*applicationv1alpha1.App, name string) (*applicationv1alpha1.App, error) {
+func getAppCR(list []*applicationv1alpha1.App, namespace, name string) (*applicationv1alpha1.App, bool) {
```
TheoBrigitte: There is a new `namespace` argument but it is never used for comparison.

kopiczko: LoL

service/controller/v1/resource/app/resource.go
```diff
@@ -66,23 +66,17 @@ func (r *Resource) Name() string {
...
 	for _, l := range list {
-		if l.Name == name {
-			return l, nil
+		b := true
+		b := b && l.Namespace == namespace
+		b := b && l.Name == name
+		if b {
```
TheoBrigitte: `if l.Namespace == namespace && l.Name == name {` would have been fine also.

kopiczko: Yeah I'm just getting ready for the future that will never happen :D

service/controller/v1/resource/app/fixture_app_cr_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"time"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	applicationv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func newTestAppCRFromFilled(modifyFunc func(*applicationv1alpha1.App)) *applicationv1alpha1.App {
```
kopiczko: This thing here follows the approach from `e2etemplates` repo.

I'm thinking about extracting top-level `fixture` package which we could easily copy around our projects.

kopiczko:
@TheoBrigitte PTALA. Tests added.

service/controller/v1/resource/app/fixture_app_cr_test.go
kopiczko: This may look like an overkill for that PR, but I also want to test `isAppCRModified` with it.

kopiczko:
Going with that. Had a chat about the changes with @TheoBrigitte.

PR #50, opened 2019-02-12
label selector

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/5273

- Use informer label selector in order to only watches ReleaseCycle CR matching the operator version.
- Moved operator version and information into `pkg/project` package.


pkg/project/version.go
```diff
@@ -0,0 +1,33 @@
+package project
+
+const (
+	notAvailable = "n/a"
+)
+
+var (
+	description = "The release-operator manages chart configs for new releases."
+	gitSHA      = notAvailable
```
xh3b4sd: We need to check if this works with `architect`. I am pretty sure it does not, because it says `-X main.gitCommit=%s`. I would also question if the variable thing works that way. I remember an issue in some other project and we had to use the plain string. I would prefer that. No need for the constant. See https://github.com/giantswarm/architect/blob/8910667e1e26464c1d6d4c8935f445afcdf03384/workflow/golang.go#L261.

kopiczko: Yes we need to change the `architect` first. It should set all those variables (including old ones) for the time being.

Note it isn't a `const`. It's a `var`.

xh3b4sd: `notAvailable` is a constant and I was questioning if the build flags even overwrite their content. I would simply go with a string.

kopiczko: Ack. +1 on that.

service/controller/release.go
```diff
@@ -53,6 +61,9 @@ func NewRelease(config ReleaseConfig) (*Release, error) {
 			Logger:  config.Logger,
 			Watcher: config.G8sClient.Release().ReleaseCycles(""),
 
+			ListOptions: metav1.ListOptions{
+				LabelSelector: fmt.Sprintf("%s=%s", releaseOperatorVersionLabel, project.Version()),
```
kopiczko: We need to support `service.controller.watch.version` setting here.

pkg/project/version.go
kopiczko:
```suggestion
	gitSHA      = "n/a"
```

TheoBrigitte: I removed the constant, and will tackle architect afterward.

service/controller/release.go
TheoBrigitte: Removed the label selector from this PR, as we spoke and it's more complex than this.

PR #53, opened 2019-02-13
service: add skeleton for release controller

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/5282.

- status resource will set status of Release CRs.
- app resource will create App CRs for release components.

### DIFF
--- a/service/controller/release.go
+++ b/service/controller/release.go
@@ -1,0 +1,107 @@
+package controller
+
+import (
+	releasev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/release/v1alpha1"
+	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/giantswarm/operatorkit/client/k8scrdclient"
+	"github.com/giantswarm/operatorkit/controller"
+	"github.com/giantswarm/operatorkit/informer"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/giantswarm/release-operator/service/controller/release"
+)
+
+type ReleaseConfig struct {
+	G8sClient    versioned.Interface
+	K8sClient    kubernetes.Interface
+	K8sExtClient apiextensionsclient.Interface
+	Logger       micrologger.Logger
+
+	ProjectName string
+}
+
+type Release struct {
+	*controller.Controller
+}
+
+func NewRelease(config ReleaseConfig) (*Release, error) {
+	if config.G8sClient == nil {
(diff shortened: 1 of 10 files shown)


service/controller/release.go
```diff
@@ -0,0 +1,107 @@
...
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	var newInformer *informer.Informer
+	{
+		c := informer.Config{
+			Logger:  config.Logger,
+			Watcher: config.G8sClient.Release().Releases(""),
```
rossf7: Is there a reason the watch namespace isn't set?

service/controller/release.go
```diff
@@ -0,0 +1,107 @@
...
+	var releaseController *controller.Controller
+	{
+		c := controller.Config{
+			CRD:       releasev1alpha1.NewReleaseCRD(),
+			CRDClient: crdClient,
+			Informer:  newInformer,
+			Logger:    config.Logger,
+			ResourceSets: []*controller.ResourceSet{
+				resourceSet,
+			},
```
xh3b4sd: Should we drop the concept of the resource set later with VOO and simply configure the resources in the controller?

service/controller/release/resource_set.go
```diff
@@ -0,0 +1,129 @@
...
+
+		appResource, err = app.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	resources := []controller.Resource{
+		statusResource,
+		appResource,
```
xh3b4sd: Can you elaborate on the app resource? I don't follow what this is doing.

service/controller/release.go
kopiczko: This is a cluster scope CRD so I think it must be like that. But anyway I think we always do that to watch resources from all namespaces.

service/controller/release.go
kopiczko: Yes it's a good call. I'll take a note

service/controller/release/resource_set.go
kopiczko: Yes. `app` resource in `release` controller will create App CR for each component in the release. E.g. it will install `aws-operator.v1.2.3`.

`app` resource in `releasecycle` controller will create App CR for release which observed ReleaseCycle CR describes. E.g. `aws.v2.3.4`.

service/controller/release.go
rossf7: Oh wow yes that makes sense since its cluster scoped. Thx!

service/controller/release/resource_set.go
TheoBrigitte: Late to the party, but about naming: how about having
- `appComponentResource` which creates components App CR
- `appReleaseResource` which creates release App CR

wdyt?

kopiczko: They will be in separate controllers so we can implement those two things in separate resources without having a name clash.

PR #55, opened 2019-02-14
Never uninstall Release CR

TheoBrigitte:

This PR ensures release-operator never uninstall `Release` CR, as mentioned in the spec :

> We never delete and mutate Release CRs


TheoBrigitte:
That's a concern we have to take in consideration. I just added it to our task list: https://github.com/giantswarm/giantswarm/issues/5273

xh3b4sd:
Let's check if there are admission controllers or webhooks for delete endpoints. That would be nice.

TheoBrigitte:
Will do this as part of the `- [] Ensure Release CR cannot be deleted` task.

My goal here is just to ensure this operator is not uninstalling `Release` CR.

kopiczko:
> What about ensuring this with the RBAC rules and not giving it delete permissions for the CR? What happens when somebody does `kubectl delete`?

Nothing that bad happens. We should recreate Release CR from ReleaseCycle CR again. Those are are tagged releases so we are guaranteed they will be the same and the system will reconcile again.

PR #57, opened 2019-02-14
add key functions to manipulate release cycle

TheoBrigitte:
Towards: [giantswarm/giantswarm#5282](https://github.com/giantswarm/giantswarm/issues/5282)

Add some key functions to manipulate release cycle CR and release information.


service/controller/key/key.go
```diff
@@ -21,10 +24,36 @@ const (
 	ServiceTypeManaged = "managed"
 )
 
-func ReleaseVersion(customResource releasev1alpha1.Release) string {
-	return customResource.Spec.Version
+// ReleaseAppCRName returns the name of the release App CR for the given release cycle.
+func ReleaseAppCRName(releaseCycleCR releasev1alpha1.ReleaseCycle) string {
```
xh3b4sd: The app CR wording confuses me. The type given is `ReleaseCycle`. What does "app" have to do with it? What app?

service/controller/key/key.go
```diff
@@ -21,10 +24,36 @@ const (
...
+func ReleaseVersion(releaseCR releasev1alpha1.Release) string {
+	return releaseCR.Spec.Version
+}
+
+// SplitReleaseName splits a release name into provider and version.
+// It returns provider, version, and error, in this order.
+//
+// It expects name to be in the following format <provider>.<version>
+// e.g. aws.v6.0.1
+func SplitReleaseName(name string) (string, string, error) {
```
xh3b4sd: We need unit tests for this. Thinking further about what this method does makes me wonder why we should derive any meta information from the name of anything. The better design would be to use the name to lookup whatever resource we want to fetch and take the relevant information from there. Provider and version should be tracked in the `Release` CR.

service/controller/key/key.go
TheoBrigitte: The idea here is: from a release cycle CR we produce the name of the app CR which is going to install the release CR.

Not sure how to express this better.

service/controller/key/key.go
TheoBrigitte: > We need unit tests for this.

I agree on this.

> The better design would be to use the name to lookup whatever resource we want to fetch and take the relevant information from there. Provider and version should be tracked in the `Release` CR.

There is no `Release` CR installed yet at the time the `ReleaseCycle` CR is created, since we use the latter to create the former. Hence we cannot fetch the information from the `Release` CR.

We also decided not to store the release name and version as separate fields in the `ReleaseCycle` CR as this open the door to store even more information there like the url of the catalog, kubeconfig and so on. Which would make `ReleaseCycle` CR be a pale copy of `App` CR.

We should be able to rely on the name here to provide us with the relevant information we need. If the name is wrong, it means that something is broken and has to be fixed.

xh3b4sd: Understood. I think that smells but I do not have a better idea now. Let's have unit tests though.

service/controller/key/key.go
kopiczko: Is it useful to expose that in the `key` package. I think it should be known only by the `app` resource in `releasecycle` controller.

service/controller/key/key.go
```diff
@@ -21,10 +24,36 @@ const (
...
-	return customResource.Spec.Version
+// ReleaseAppCRName returns the name of the release App CR for the given release cycle.
+func ReleaseAppCRName(releaseCycleCR releasev1alpha1.ReleaseCycle) string {
+	return ReleasePrefix(releaseCycleCR.GetName())
+}
+
+// ReleasePrefix adds release- prefix to name.
+func ReleasePrefix(name string) string {
+	return fmt.Sprintf("release-%s", name)
+}
```
kopiczko: Same for this. I don't think it's useful to expose that in the key package. We should have stuff shared between resources in here.

service/controller/key/key.go
TheoBrigitte: make sense. I moved the function to the `app` resource.

service/controller/key/key.go
TheoBrigitte: added unit test. PTAL

service/controller/key/key.go
TheoBrigitte: move this function to `app` resource.

TheoBrigitte:
fishing for approval

service/controller/key/key_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}{
+		{
+			name:             "case 0: valid semver",
+			input:            "aws.v6.0.1",
+			expectedProvider: "aws",
+			expectedVersion:  "v6.0.1",
+			errorMatcher:     nil,
+		},
+		{
+			name:             "case 0: valid any version",
```
rossf7:
```suggestion
			name:             "case 1: valid any version",
```

Minor but test cases need updating.

service/controller/releasecycle/resource/app/resource.go
```diff
@@ -87,3 +90,13 @@ func toAppCRs(v interface{}) ([]*applicationv1alpha1.App, error) {
 
 	return appCRs, nil
 }
+
+// ReleaseAppCRName returns the name of the release App CR for the given release cycle.
```
rossf7:
```suggestion
// releaseAppCRName returns the name of the release App CR for the given release cycle.
```

service/controller/releasecycle/resource/app/resource.go
```diff
@@ -87,3 +90,13 @@ func toAppCRs(v interface{}) ([]*applicationv1alpha1.App, error) {
...
+func releaseAppCRName(releaseCycleCR releasev1alpha1.ReleaseCycle) string {
+	return releasePrefix(releaseCycleCR.GetName())
+}
+
+// ReleasePrefix adds release- prefix to name.
```
rossf7:
```suggestion
// releasePrefix adds release- prefix to name.
```

service/controller/releasecycle/resource/app/resource.go
rossf7: Yes in the past we've stuffed too much stuff in the key package. I'll keep this in mind for app-operator too.

PR #60, opened 2019-02-15
main, project, service: use project package

kopiczko:


service/controller/release_cycle.go
```diff
@@ -11,16 +11,19 @@ import (
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/client-go/kubernetes"
 
+	"github.com/giantswarm/release-operator/pkg/project"
 	"github.com/giantswarm/release-operator/service/controller/releasecycle"
 )
 
+var (
+	releaseCycleControllerName = project.Name() + "-releasecycle"
+)
```
xh3b4sd: `var` is maybe weird. `const` for a format string would be safer. Some nice godoc would then also be useful to show how it looks like.

kopiczko: `const` can't be used because `project.Name()` is dynamic.

What do you mean by godoc? Describing that the value is `release-opreator-releasecycle`? I think it's rather obvious. But I can add if you insist.

I put it on the top of the file because we should change that value when copy/pasting controllers.

xh3b4sd: I suggested to have a constant for a format string. 

```
const (
    // releaseCycleControllerFormat is the format used to compute the operatorkit controller 
    // name. The computed string looks like follows.
    //
    //      release-operator-releasecycle
    //
    releaseCycleControllerFormat = "%s-releasecycle"
)
```

kopiczko: I prefer to stick with a `var`. I really dislike having format string somewhere and Printf call somewhere else. `go vet` becomes useless which I find a bigger risk than having a var. I don't think we'll ever run into issues because of having a var and it's straightforward to read but if you feel unsafe I can convert that to a function.

PR #61, opened 2019-02-20
release app resource

TheoBrigitte:

This PR update the app resource in order to have it create release App CR for every corresponding ReleaseCycle CR.


service/controller/releasecycle/resource/app/create.go
```diff
@@ -2,62 +2,61 @@ package app
...
+	r.logger.LogCtx(ctx, "level", "debug", "message", "computing create state", "app", desiredAppCR.GetName())
 
-		for _, d := range desiredAppCRs {
-			_, ok := getAppCR(currentAppCRs, d.Namespace, d.Name)
-			if !ok {
-				appCRsToCreate = append(appCRsToCreate, d)
-			}
+	var createAppCR *applicationv1alpha1.App
+	{
+		if currentAppCR.GetName() == "" {
```
xh3b4sd: Based on `GetCurrentState`, this code here will panic.

service/controller/releasecycle/resource/app/desired.go
```diff
@@ -4,84 +4,56 @@ import (
...
+	return appCR, nil
 }
 
-func (r *Resource) newAppCR(ctx context.Context, releaseCR releasev1alpha1.Release, component releasev1alpha1.ReleaseSpecComponent) (*applicationv1alpha1.App, error) {
+func (r *Resource) newAppCR(name, repository, version string) *applicationv1alpha1.App {
 	appCR := &applicationv1alpha1.App{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "App",
 			APIVersion: "application.giantswarm.io",
 		},
```
xh3b4sd: We got a handy helper for this. See https://github.com/giantswarm/apiextensions/blob/master/pkg/apis/application/v1alpha1/app_types.go#L53-L58.

service/controller/releasecycle/resource/app/create.go
TheoBrigitte: This does not panic since `GetCurrentState` always return an empty `applicationv1alpha1.App{}` object.

see https://github.com/giantswarm/apiextensions/blob/master/pkg/clientset/versioned/typed/application/v1alpha1/app.go#L68

Should I also add a test for nil in case something change there ?

service/controller/releasecycle/resource/app/desired.go
TheoBrigitte: Nice

service/controller/releasecycle/resource/app/create.go
xh3b4sd: That is a pretty dangerous assumption. The interface returns `nil`, so when no object could be found, `nil` should be expected, regardless any implementation detail. Let's be defensive. I would rather check for `nil` here additionally in the first place just to be safe. A unit test for this is also a good shout.

service/controller/releasecycle/resource/app/create.go
```diff
@@ -2,62 +2,61 @@ package app
...
-	"fmt"
 
 	applicationv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/microerror"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	"github.com/giantswarm/release-operator/service/controller/key"
 )
 
+// ApplyCreateChange ensures createChange App CR is created in k8s api.
```
rossf7:
```suggestion
// ApplyCreateChange ensures the App CR is created in the k8s api.
```

I'd actually be fine without the godoc here as this is standard operatorkit behaviour. I usually just add godoc for the resource constructor.

service/controller/releasecycle/resource/app/create.go
```diff
@@ -2,62 +2,61 @@ package app
...
-		r.logger.LogCtx(ctx, "level", "debug", "message", "App CRs do not have to be created in the Kubernetes API")
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensured creation of release App CR", "app", appCR.GetName())
 	}
 
 	return nil
 }
 
+// newCreateChange computes the App CR to be created.
+//
+// nil, nil is returned when no App CR has to be created.
```
rossf7: I'd say this computes whether the CR should be created. I'd say the app CR itself is computed by the desired state. Again I'd actually be fine without the godoc here.

service/controller/releasecycle/resource/app/create.go
TheoBrigitte: I added check for `nil` and unit tests. PTAL

service/controller/releasecycle/resource/app/create.go
```diff
@@ -2,62 +2,61 @@ package app
...
-	appCRs, err := toAppCRs(createChange)
+	appCR, err := key.ToAppCR(createChange)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
-	if len(appCRs) != 0 {
-		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating %d App CRs in the Kubernetes API", len(appCRs)))
+	if appCR != nil {
+		r.logger.LogCtx(ctx, "level", "debug", "message", "ensuring creation of release App CR", "app", appCR.GetName())
```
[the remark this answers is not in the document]
TheoBrigitte: I felt like using an extra arg would make the log clearer. But I can change it to the old `fmt.Sprintf` style.

service/controller/releasecycle/resource/app/create.go
TheoBrigitte: godoc make things explicit here. I updated it.

service/controller/releasecycle/resource/app/create.go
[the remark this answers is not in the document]
rossf7: Yes code wise I like it as its cleaner.

My only worry is if it causes any issues ingesting the logs. For example we separate out `"stack"` and the stack traces. I don't understand our EFK stack well enough to say on that.

service/controller/releasecycle/resource/app/resource.go
```diff
@@ -27,8 +26,13 @@ type Config struct {
 }
 
 // Resource implements the app resource.
+//
+// It ensures each release cycle has its corresponding release installed.
+// It does so by creating an App CR for the release, which will then be
+// installed by app-operator.
+// Note: release are never removed, so removing a release cycle CR has no effect
```
rossf7:
```suggestion
// Note: releases are never removed, so removing a release cycle CR has no effect
```

service/controller/releasecycle/resource/app/create.go
```diff
@@ -2,62 +2,61 @@ package app
...
+	desiredAppCR, err := key.ToAppCR(desiredState)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
 
-	var appCRsToCreate []*applicationv1alpha1.App
-	{
-		r.logger.LogCtx(ctx, "level", "debug", "message", "computing create state")
+	var createAppCR *applicationv1alpha1.App
+	if desiredAppCR != nil {
```
xh3b4sd: Desired should never be `nil` and we have always the control over that, don't we?

service/controller/releasecycle/resource/app/create.go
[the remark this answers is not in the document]
TheoBrigitte: That's a good point. I reverted that to the old style logging with `fmt.Sprintf` then.

service/controller/releasecycle/resource/app/create.go
TheoBrigitte: Yes but I'd rather have this `nil` check here which guarantee that stuff don't blow up when desired can be nil. I'd rather be extra safe. Going with this.

PR #62, opened 2019-02-22
add --release.appcatalog flag

TheoBrigitte:

Provide `--release.appcatalog flag` argument in order to configure the catalog CR name to install releases from.

### DIFF
--- a/main.go
+++ b/main.go
@@ -102,6 +102,7 @@ func mainWithError() (err error) {
 
 	daemonCommand := newCommand.DaemonCommand().CobraCommand()
 
+	daemonCommand.PersistentFlags().String(f.Release.AppCatalog, "", "application catalog name. Used to install releases.")
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.Address, "", "Address used to connect to Kubernetes. When empty in-cluster config is created.")
 	daemonCommand.PersistentFlags().Bool(f.Service.Kubernetes.InCluster, true, "Whether to use the in-cluster config to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Service.Kubernetes.TLS.CAFile, "", "Certificate authority file path to use to authenticate with Kubernetes.")
--- a/service/controller/releasecycle/resource_set.go
+++ b/service/controller/releasecycle/resource_set.go
@@ -24,6 +24,7 @@ type ResourceSetConfig struct {
 	Logger    micrologger.Logger
 
 	// Settings.
+	AppCatalog  string
 	ProjectName string
 }
 
@@ -37,7 +38,8 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 			K8sClient: config.K8sClient,
 			Logger:    config.Logger,
 
-			Namespace: "giantswarm",
+			AppCatalog: config.AppCatalog,
+			Namespace:  "giantswarm",
 		}
 
 		ops, err := app.New(c)
--- a/flag/flag.go
+++ b/flag/flag.go
@@ -3,11 +3,13 @@ package flag
 import (
 	"github.com/giantswarm/microkit/flag"
 
+	"github.com/giantswarm/release-operator/flag/release"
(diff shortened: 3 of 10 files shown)


main.go
```diff
@@ -102,6 +102,7 @@ func mainWithError() (err error) {
 
 	daemonCommand := newCommand.DaemonCommand().CobraCommand()
 
+	daemonCommand.PersistentFlags().String(f.Release.AppCatalog, "", "application catalog name. Used to install releases.")
```
xh3b4sd: The catalogue is configured to the release cycle controller. The flag should then reflect it, shouldn't it?

TheoBrigitte: Yes this flag configures the release cycle controller, but it sets up the value for the release app catalog. So I found it more accurate to have `--release.appcatalog` than `--releasecycle.appcatalog`

service/controller/releasecycle/resource_set.go
```diff
@@ -37,7 +38,8 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 			K8sClient: config.K8sClient,
 			Logger:    config.Logger,
 
-			Namespace: "giantswarm",
+			AppCatalog: config.AppCatalog,
```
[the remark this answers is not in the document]
TheoBrigitte: All release apps will be available in the [control-plane catalog](https://github.com/giantswarm/control-plane-catalog), which will be installed by `opsctl ensure appcatalogs`.

So the catalog we install releases from is the same for all releases.

main.go
rossf7: I agree. If I'm understanding things correctly the app CRs being created are so the releases get installed. So the naming reflecting that is clearer for me.

service/controller/releasecycle/resource_set.go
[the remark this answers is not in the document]
rossf7: Using a flag makes sense to me. Since the Helm charts for the apps are always being pulled from the `control-plane` catalog.

main.go
xh3b4sd: I have still difficulties getting the mental model right. Do we have one picture that shows how releases, apps and catalogues work together and what is meant for what?

xh3b4sd: So there is an app catalogue. The catalogue contains apps. Where are releases? Is there a catalogue per release? Or is there a release per app?

rossf7: I'm not super deep into the details but both specs have diagrams and these are the best we have AFAIK.

https://github.com/giantswarm/giantswarm/blob/master/specs/versioning-of-operators.md
https://github.com/giantswarm/giantswarm/blob/master/roadmap/MANAGED-SERVICES-CATALOG.MD

Physically a catalog is just somewhere we store tarballs of Helm charts. These are the unit of deployment. You can see how the test app in the sample catalog is versioned.

https://github.com/giantswarm/sample-catalog

We won't have many catalogs. AIUI the `control-plane-catalog` is where the releases will be stored.

xh3b4sd: Checking the VOO spec does not properly explain to me what an app catalog is and how it is involved. I miss this somehow. The MS spec tells me that a catalog contains release CRs, which is weird because I would naively assume that the app catalog contains, well, apps. So this is still not clear. There are four components. Release Cycle, Release, App Catalog, App. I would like to have a simple picture that shows how these 4 things relate to each other. This is the missing piece for me.

TheoBrigitte: A release is represented by a Release CR which is then packaged as a Helm chart in the `control-plane-catalog`. This is described here: https://github.com/giantswarm/giantswarm/blob/master/specs/versioning-of-operators.md#giantswarmrelease-provider-repository

Later on this release is installed on the cluster by `release-operator` which create an `App` CR referencing the release in the `control-plane-catalog. This is described here: https://github.com/giantswarm/giantswarm/blob/master/specs/versioning-of-operators.md#release-operator

An improvement to the latter diagram could be to add more details to the App CR, to show that the release is taken from the `control-plane-catalog`. Would this help ?

xh3b4sd: Yeah maybe. It is confusing that it is basically catalog -> release -> app. The relation is not intuitive.

PR #65, opened 2019-02-27
pkg/controller: create generic app resource

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/5282
Towards https://github.com/giantswarm/giantswarm/issues/5461

This resource is meant to be used by two controllers. There are only
current and desired state functions to be defined.

Note that in this implementation GetDesiredStateFunc should be aware if
the object is created/updated or deleted. That is easy to achieve with
key.IsAppDeleted.

### DIFF
--- a/pkg/controller/resource/app/resource.go
+++ b/pkg/controller/resource/app/resource.go
@@ -1,0 +1,87 @@
+package app
+
+import (
+	"context"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// Name is the identifier of the resource.
+	Name = "app"
+)
+
+type Config struct {
+	G8sClient versioned.Interface
+	K8sClient kubernetes.Interface
+	Logger    micrologger.Logger
+
+	GetCurrentStateFunc func(ctx context.Context, obj interface{}) ([]*v1alpha1.App, error)
+	GetDesiredStateFunc func(ctx context.Context, obj interface{}) ([]*v1alpha1.App, error)
+}
+
+type Resource struct {
+	g8sClient versioned.Interface
+	k8sClient kubernetes.Interface
+	logger    micrologger.Logger
+
+	getCurrentStateFunc func(ctx context.Context, obj interface{}) ([]*v1alpha1.App, error)
+	getDesiredStateFunc func(ctx context.Context, obj interface{}) ([]*v1alpha1.App, error)
+}
+
+func New(config Config) (*Resource, error) {
+	if config.G8sClient == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.G8sClient must not be empty", config)
+	}
+	if config.K8sClient == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.K8sClient must not be empty", config)
+	}
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
+	}
+
+	if config.GetCurrentStateFunc == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.GetCurrentStateFunc must not be empty", config)
+	}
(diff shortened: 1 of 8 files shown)


pkg/controller/resource/app/resource.go
```diff
@@ -0,0 +1,87 @@
...
+	for _, a := range crs {
+		if cr.Name == a.Name && cr.Namespace == a.Namespace {
+			return true
+		}
+	}
+
+	return false
+}
+
+func toState(v interface{}) ([]*v1alpha1.App, error) {
```
rossf7: I find this name a bit confusing. It implies to me that there is a State struct whereas its returning app CRs.

What about `toAppCRs`?

pkg/controller/resource/app/resource.go
```diff
@@ -0,0 +1,87 @@
...
+	Name = "app"
+)
+
+type Config struct {
+	G8sClient versioned.Interface
+	K8sClient kubernetes.Interface
+	Logger    micrologger.Logger
+
+	GetCurrentStateFunc func(ctx context.Context, obj interface{}) ([]*v1alpha1.App, error)
+	GetDesiredStateFunc func(ctx context.Context, obj interface{}) ([]*v1alpha1.App, error)
```
xh3b4sd: I would find it cooler to have an interface of a partial resource that the other side implements and that is injected here.

pkg/controller/resource/app/update.go
```diff
@@ -0,0 +1,103 @@
...
+	merged := current.DeepCopy()
+
+	merged.Annotations = desired.Annotations
+	merged.Labels = desired.Labels
+
+	merged.Spec = desired.Spec
+
+	if reflect.DeepEqual(current, merged) {
+		return nil
+	}
```
rossf7: I really like how the comparison is done here.

pkg/controller/resource/app/resource.go
kopiczko: Maybe it makes sense then to establish a practice to always have a `State` struct? I like having `desired.AppCRs`. There would also be less guessing what the actual state type is.

pkg/controller/resource/app/spec.go
```diff
@@ -0,0 +1,23 @@
+package app
+
+import (
+	"context"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+)
+
+type StateGetter interface {
```
kopiczko: I don't think it's the best name. Suggestions welcome.

pkg/controller/resource/app/resource.go
kopiczko: I extracted an interface. PTALA.

pkg/controller/resource/app/resource.go
rossf7: desired.AppCRs is nice. Its very clear what the type it. So this would be wrapped in a `State` struct? I like that as an approach.

pkg/controller/resource/app/spec.go
rossf7: What about just `Interface`? We do that in `helmclient` and a few other places.

kopiczko: Interface is ok, but should I name field `Resource.interface` then? Is it even allowed?

rossf7: Ack yes Resource.interface isn't nice even if it worked. What about `StateInterface`? 

Its the Getter part of the name I'm not keen on because I don't think we use it in the rest of our codebase.

Another thought. Should the Name be part of the interface or just a setting?

kopiczko: Checked `Resource.interface` doesn't work. I think `StateGetter` goes more into conventions like `Reader` and `Closer`. Especially when `Name` is moved as a resource setting which I think is a good shout.

rossf7: 👍 Cool, `StateGetter` makes sense to me.

pkg/controller/resource/app/resource.go
kopiczko: I changed that to `toAppCRs` for now. I'll introduce `State` struct in a separate PR so we can discuss that separately. I also thought that if we have `State` we should also have `Change`?

kopiczko:
Merging to keep me going. I think it's good enough. I think returning slice of Apps makes sense for the current/desired state for now. I want to keep going. We can introduce that later if needed.

PR #72, opened 2019-03-05
clean up releasecycle controller

TheoBrigitte:

Remove unused elements:
- configmap resource
- secret resource
- controllercontext


TheoBrigitte:
> So why don't we need this anymore?

The releasecycle controller, only takes care of creating releases App CR which do not require any specific configmap or secret. This was just legacy from the early times of release-operator which was deploying chartconfigs.

PR #76, opened 2019-03-07
release/resource/app: initial implementation

kopiczko:

This resource is responsible for creating/updating App CRs for components in non-EOL
releases and deleting App CRs for components in EOL releases. App CRs for
components existing in both EOL and non-EOL releases are not deleted by this resource.


service/controller/release/resource/app/resource.go
```diff
@@ -36,6 +39,8 @@ func New(config Config) (*Resource, error) {
 		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
 	}
 
+	c := app.Config{}
+
```
xh3b4sd: That should not compile.

service/controller/release/resource/app/resource.go
```diff
@@ -56,3 +61,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
```
xh3b4sd: What about these methods here?

service/controller/release/resource/app/desired.go
```diff
@@ -0,0 +1,127 @@
...
+	cr.GetDeletionTimestamp()
+	if cr.Status.Cycle.Phase != releasev1alpha1.CyclePhaseEOL && !key.IsDeleted(cr) {
+		return cr.Spec.Components, nil
+	}
+
+	// TODO unit-test long path.
+	// TODO This is very suboptimal and likely to cause problems in the
+	// future. All Release CRs CRs are listed every time for all EOL
+	// releases. More EOL releases we have this becomes worse. We should
+	// optimize it with some sort of caching or labelling.
```
xh3b4sd: Labels should be the way to go here.

kopiczko:
> I am not sure if I should review this already. 🙈

This is a draft. Many peaces are left. I just wanted to show direction and get eventual feedback if I'm doing something stupid.

service/controller/release/resource/app/resource.go
```diff
@@ -15,44 +16,61 @@ const (
...
 
-type Resource struct {
-	g8sClient versioned.Interface
-	k8sClient kubernetes.Interface
-	logger    micrologger.Logger
+	AppCatalog string
 }
 
-func New(config Config) (*Resource, error) {
+func New(config Config) (controller.Resource, error) {
```
kopiczko: The diff here is ugly. I recommend looking at the file.

kopiczko:
I consider this ready for review. I only want to add unit tests for `getDesiredComponents` (there are TODOs in the code).

service/controller/release/resource/app/desired.go
```diff
@@ -0,0 +1,121 @@
...
+			return nil, microerror.Mask(err)
+		}
+
+		activeReleases = result.Items
+	}
+
+	var desiredComponents []releasev1alpha1.ReleaseSpecComponent
+
+	// If this an EOL release only components that are also part of other
+	// non-EOL (active) releases are desired.
```
xh3b4sd: Why? What is usually obviously expressed by the code, the why is not. So let's please write down the cause.

service/controller/release/resource/app/desired.go
```diff
@@ -0,0 +1,121 @@
...
+	// components of that release are desired.
+	cr.GetDeletionTimestamp()
+	if cr.Status.Cycle.Phase != releasev1alpha1.CyclePhaseEOL && !key.IsDeleted(cr) {
+		return cr.Spec.Components, nil
+	}
+
+	var activeReleases []releasev1alpha1.Release
+	{
+		opts := metav1.ListOptions{
+			LabelSelector: key.LabelReleaseCyclePhase + "!=" + releasev1alpha1.CyclePhaseEOL.String(),
```
xh3b4sd: Is there another more standard way of computing the label selector?

service/controller/release/resource/app/desired.go
```diff
@@ -0,0 +1,121 @@
...
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("computed %d desired app CRs for release %#q components", len(desiredAppCRs), cr.Name))
+	}
+
+	return desiredAppCRs, nil
+}
+
+func (r *resourceStateGetter) getDesiredComponents(ctx context.Context, cr *releasev1alpha1.Release) ([]releasev1alpha1.ReleaseSpecComponent, error) {
+	// If this is a non-EOL release and it isn't deletion event all
+	// components of that release are desired.
+	cr.GetDeletionTimestamp()
```
xh3b4sd: That code is not doing anything IMO.

kopiczko: Yeah it's a leftover.
```suggestion
```

service/controller/release/resource/app/desired.go
kopiczko: I don't know. Everything I saw so far was string concatenation or formatting.

service/controller/release/resource/app/desired.go
kopiczko: Would something like that work?

> If this an EOL release components that are also part of other non-EOL (active) releases are still desired. Otherwise active releases consisting of those shared components would stop working till next reconciliation loop.

xh3b4sd: That makes sense to me up to the point of the next reconciliation loop. That opens more questions. But yeah, let'S extend the comment a bit and go.

kopiczko: Then maybe:

> If all App CRs for components of this EOL release were blindly deleted there is a chance some active (non-EOL) release sharing some of those components would stop working (that would be eventually fixed when the broken release is reconciled again and App CRs for missing components are recreated). To avoid such situations this release components that are also part of other active releases are still desired.

xh3b4sd: `situations this release` -> `situations these release`

kopiczko: `To avoid such situations this release components that are also part of other active releases are still desired`

`this release components` - meaning the components of the (this) Release CR that is already reconciled. I meant `this release` not `these ... components`. Maybe `this release's components` is more accurate.

xh3b4sd: Then it is `this release's components` IMO.

kopiczko: Done.

service/controller/release/resource/app/desired.go
```diff
@@ -0,0 +1,125 @@
...
+		opts := metav1.ListOptions{
+			LabelSelector: key.LabelReleaseCyclePhase + "!=" + releasev1alpha1.CyclePhaseEOL.String(),
+		}
+
+		result, err := r.g8sClient.ReleaseV1alpha1().Releases("").List(opts)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		activeReleases = result.Items
```
TheoBrigitte: `!key.IsDeleted(cr)` is used earlier to filter out deleted releases. I think the same logic is missing here in order to filter out deleted releases from `activeReleases`.

kopiczko: Yeah that's a good catch. I should add unit tests with CRs with deletion timestamp set.

kopiczko: Done together with tests. PTALA.

service/controller/release/resource/app/desired.go
```diff
@@ -70,7 +70,13 @@ func (r *resourceStateGetter) getDesiredComponents(ctx context.Context, cr *rele
 			return nil, microerror.Mask(err)
 		}
 
-		activeReleases = result.Items
+		for _, r := range result.Items {
+			if key.IsDeleted(&r) {
```
tomahawk28: We can make it simple as
```
			if !key.IsDeleted(&r) {
				activeReleases = append(activeReleases, r)
			} 			
```

kopiczko: Yeah I still prefer current approach. It makes diffs prettier if we add more conditions. I also like _early returns_ to the point it's ridiculous.

kopiczko:
Tested it manually. Works nice.

I fixed what @TheoBrigitte requested to change. He is on vacation so I will proceed w/o waiting for re-review.

E2E tests should come this week.

PR #77, opened 2019-03-07
key, release: add label resource

kopiczko:

This resource sets "release-operator.giantswarm.io/release-cycle-phase"
on the observed CR. The value is taken from .Status.Cycle.Phase. This is
to allow selecting non-EOL Release CRs efficiently using label selector.


service/controller/key/key.go
```diff
@@ -14,6 +14,7 @@ const (
 
 	LabelAppOperatorVersion = "app-operator.giantswarm.io/version"
 	LabelManagedBy          = "giantswarm.io/managed-by"
+	LabelReleaseCyclePhase  = "release-operator.giantswarm.io/cycle-phase"
```
xh3b4sd:
```suggestion
	LabelReleaseCyclePhase  = "release-operator.giantswarm.io/release-cycle-phase"
```

service/controller/release/resource/label/create.go
```diff
@@ -0,0 +1,70 @@
...
+		return microerror.Mask(err)
+	}
+
+	{
+		r.logger.LogCtx(ctx, "level", "debug", "message", "finding if custom resource needs to be updated")
+
+		currentLabels := cr.Labels
+		if currentLabels == nil {
+			currentLabels = map[string]string{}
+		}
```
xh3b4sd: That does not seem to make sense. What is `currentLabels` good for?

service/controller/release/resource/label/create.go
```diff
@@ -0,0 +1,70 @@
...
+
+		v := cr.Labels[key.LabelReleaseCyclePhase]
```
xh3b4sd: I think we should check if the label even exists. Above you just initialized the label map.

service/controller/release/resource/label/resource.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	Name = "status"
```
xh3b4sd: I think this is the `label` resource.

service/controller/release/resource/label/create.go
kopiczko: Yeah this should be:

```suggestion
		v := currentLabels[key.LabelReleaseCyclePhase]
```

Zero value of the string is `""` so no problem with value not existing in the map.

service/controller/release/resource/label/create.go
kopiczko: The code below got updated. They are here to make sure we not try to get value from `nil` map.

service/controller/release/resource/label/resource.go
kopiczko:
```suggestion
	Name = "label"
```

kopiczko:
PTALA

service/controller/release/resource/label/create.go
xh3b4sd: I am still wondering why we use a new variable instead of simply `cr.Labels`. That is even shorter than the variable you chose and the value is a map anyway so copying it does not do something different.

kopiczko: Are you sure `cr.Labels` can't be null? I don't want to mutate the CR here.

service/controller/release/resource/label/create.go
```diff
@@ -0,0 +1,70 @@
+package status
```
rossf7:
```suggestion
package label
```

This looks wrong.

service/controller/release/resource/label/create.go
xh3b4sd: It can, but why not just this? 

```
		if cr.Labels == nil {
			cr.Labels = map[string]string{}
		}
```

service/controller/release/resource/label/create.go
```diff
@@ -0,0 +1,70 @@
...
+		cr.Labels = desiredLabels
+
+		_, err = r.g8sClient.ReleaseV1alpha1().Releases(cr.GetNamespace()).Update(cr)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", "updated custom resource")
+		r.logger.LogCtx(ctx, "level", "debug", "message", "canceling reconciliation")
+		reconciliationcanceledcontext.SetCanceled(ctx)
```
rossf7: Why is the reconciliation canceled? I don't follow that part.

service/controller/release/resource/label/create.go
kopiczko: It will change the `cr` for other resources as well. Also I think the object from informer cache should not be mutated. I remember I saw that in some godoc.

Every time we update the object or its status we fetch the new instance directly from the API. I think the watched object shouldn't be mutated.

service/controller/release/resource/label/create.go
kopiczko: The object is updated so there will be next update event immediately. This is the same thing we do when updating status.

service/controller/release/resource/label/create.go
kopiczko: Thanks. I should also wire the resource.

service/controller/release/resource/label/create.go
rossf7: OK cool that makes sense.

service/controller/release/resource/label/create.go
xh3b4sd: Working on the variable changes the CR as well. Because it is a map. That is a pointer. You do not copy it with a new CR. And then, we could just copy the CR and work with it in the resource. We could even only give a copy to all resources in `operatorkit` by default.

kopiczko: Map is a value that can be `nil`. It isn't a pointer. See https://play.golang.org/p/KUhyoJ9DiCa.

service/controller/release/resource/label/create.go
kopiczko: Updated the package name and wired the resource.

service/controller/release/resource/label/create.go
xh3b4sd: `nil` is not a pointer, but a map is. When labels are `nil` and we initialize them, it does not have an effect. When labels is not `nil` and we change it, we change it for everyone, because a map is a pointer. 

* https://play.golang.org/p/tCO7P-APIO1
* https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it

kopiczko: TIL. I always thought map is a value holding a reference. Still there are no writes to the map so the CR remains unchanged. But I'll simply replace that with:

```go
		var v string
		if cr.Labels != nil {
			v = cr.Labels[key.LabelReleaseCyclePhase]
		}
```

service/controller/release/resource/label/resource.go
```diff
@@ -0,0 +1,50 @@
...
+	Logger    micrologger.Logger
+}
+
+// Resource sets "release-operator.giantswarm.io/release-cycle-phase" on the
+// observed CR. The value is taken from .Status.Cycle.Phase. This is to allow
+// selecting non-EOL Release CRs efficiently using label selector.
+//
+// Label resource and status resource are separated to be able to cancel
+// reconciliation independently. There are two requests one to update stats and
+// second to update labels which generate separate events.
```
TheoBrigitte: Love this comment <3

service/controller/release/resource/label/create.go
xh3b4sd: > We could even only give a copy to all resources in `operatorkit` by default.

What do you think about that?

service/controller/release/resource/label/resource.go
kopiczko: This is a result of todays standup :)

service/controller/release/resource/label/create.go
kopiczko: I think it's a good idea. This way we are guaranteed we don't mess with cached objects.

xh3b4sd: I made a note in my board.

PR #84, opened 2019-04-08
integration/env, integration/setup: update env vars

kopiczko:

Deviations from aws-operator:

- `E2E_` prefixes for e2e specific variables. This is to reduce mess in circle.
- `E2E_TEST_DIR` is required.
- No version bundle stuff.
- No wip/cur.


integration/env/common.go
```diff
@@ -3,57 +3,72 @@ package env
...
 	registryPullSecret = os.Getenv(EnvVarRegistryPullSecret)
 	if registryPullSecret == "" {
 		panic(fmt.Sprintf("env var '%s' must not be empty", EnvVarRegistryPullSecret))
 	}
 }
 
-func CircleCI() string {
-	return circleCI
+func CircleCI() bool {
+	return circleCI == strings.ToLower("true")
```
xh3b4sd: Shouldn't we rather use https://golang.org/pkg/strconv/#ParseBool? Also, `true` is already "to lower".

kopiczko: Good call. I didn't know it exists. Also it should be `strings.ToLower(circleCI)` LoL.

kopiczko: Problem with using strconv.ParseBool is it will fail on empty string and we need to handle an error.

kopiczko:
```suggestion
	return strings.ToLower(circleCI) =="true"
```

PR #89, opened 2019-04-12
integration: bring up tests setup to the latest advancements

kopiczko:


integration/setup/setup.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Create namespace.
+	{
+		err := config.K8sSetup.EnsureNamespaceCreated(key.Namespace)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		if !env.CircleCI() && !env.KeepResources() {
+			defer func() {
```
kopiczko: Note `defer` is used instead of usual `v == 1` madness.

integration/setup/setup.go
```diff
@@ -0,0 +1,96 @@
...
+		err = config.Release.EnsureInstalled(ctx, releaseName, chartInfo, values, installConditions...)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		if !env.CircleCI() && !env.KeepResources() {
+			defer func() {
+				deleteConditions := []release.ConditionFunc{
+					config.Release.Condition().CRDNotFound(ctx, releasev1alpha1.NewReleaseCRD()),
+					config.Release.Condition().CRDNotFound(ctx, releasev1alpha1.NewReleaseCycleCRD()),
```
kopiczko: Here is why I added `CRDNotFound` to the `e2e-harness`.

kopiczko: I actually need to get rid of that :D CRD is not deleted with the operator.

Gopkg.toml
```diff
@@ -40,7 +40,7 @@ required = [
 
 [[constraint]]
   branch = "master"
-  name = "github.com/giantswarm/e2e-harness"
+  name = "github.com/giantswarm/e2esetup"
```
xh3b4sd: NICE

kopiczko: Hold your horses :) We still have e2e-harness dependency here. But it's only one const.

kopiczko: And whole `pkg/release` package which we are moving in scope of releng SWH...

kopiczko:
This shit works. Finally. Merging.

PR #93, opened 2019-04-12
service: clean up AppCatalog configuration leftovers

kopiczko:

Follow up of https://github.com/giantswarm/release-operator/pull/91.


kopiczko:
Need to proceed here because of failing deployments.

PR #99, opened 2019-05-03
integration/setup: create App CRD in the test setup

kopiczko:

### DIFF
--- a/integration/setup/setup.go
+++ b/integration/setup/setup.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"testing"
 
+	applicationv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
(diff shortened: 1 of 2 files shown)


integration/setup/setup.go
```diff
@@ -43,7 +44,26 @@ func setup(ctx context.Context, m *testing.M, config Config) (int, error) {
...
+	// Create App CRD
+	{
+		crd := applicationv1alpha1.NewAppCRD()
+
+		err := config.K8sSetup.EnsureCRDCreated(ctx, crd)
+		if err != nil {
+			return 0, microerror.Mask(err)
+		}
+
+		if !env.CircleCI() && !env.KeepResources() {
```
JosephSalisbury: Why do we need to check if we're in circle? Because the whole thing gets torn down then anyway?

kopiczko: Yes.

PR #102, opened 2019-05-03
helm: fix rbac rules

kopiczko:

Watched object changed since we took over from spirit. Now even spirit doesn't exist anymore.


kopiczko:
Tested with integration tests.

PR #104, opened 2019-05-06
add structure for TestReleaseAppCRCreate

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/5278

Add structure for the upcoming e2e test `TestReleaseAppCRCreate`.


integration/test/releasecycle/main_test.go
```diff
@@ -0,0 +1,42 @@
+// +build k8srequired
+
+package releasecycle
```
kopiczko: Let's rename that to `releasecyclehandling` or rename `releasehandling` to `release`. I named the `releasehandling` like that because I found `release` directory weird on its own.

TheoBrigitte: I like to stick with `releasecycle` and `release`, it makes it clear and explicit what we are testing in here.

kopiczko: I still find `integration/test/release` weird. But ok with me. Let's just make it consistent.

PR #105, opened 2019-05-06
do not update release app CRs

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/5278

All the values in the release App CR which comes from the release cycle CR derivates from its name, which is by definition immutable. Therefore there is no point in having an update mechanism for release App CRs.

https://github.com/giantswarm/release-operator/blob/9b8348ca62040a8e5467cf2212fd50fcca714c2c/service/controller/releasecycle/resource/app/desired.go#L22-L31


TheoBrigitte:
> I think we should switch to the shared resource here (the same release app resource uses under the hood).

If you talk about the app resource created from `newAppCR` I totally agree but this is not part of this PR.

kopiczko:
> > I think we should switch to the shared resource here (the same release app resource uses under the hood).
> 
> If you talk about the app resource created from `newAppCR` I totally agree but this is not part of this PR.

I'm talking about https://github.com/giantswarm/release-operator/tree/master/pkg/controller/resource/app.

TheoBrigitte:
> > > I think we should switch to the shared resource here (the same release app resource uses under the hood).
> > 
> > 
> > If you talk about the app resource created from `newAppCR` I totally agree but this is not part of this PR.
> 
> I'm talking about https://github.com/giantswarm/release-operator/tree/master/pkg/controller/resource/app.

Ok, but let's make this change in another PR. Here is removed the update logic which does not make sense in the end.

TheoBrigitte:
Argument to keep this: it keeps the created resource in place sane against manual edits. Keeping it

PR #106, opened 2019-05-07
multireleasehandling: create test skeleton

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/5278.

PR #107, opened 2019-05-07
multireleasehandling: implement the first part of the test

kopiczko:

Please review #106 first.


integration/test/multireleasehandling/multi_release_handling_test.go
```diff
@@ -0,0 +1,132 @@
...
+				"cert-operator.0.1.0",
+			}
+
+			if !reflect.DeepEqual(appCRNames, expectedAppCRNames) {
+				return microerror.Maskf(waitError, "\n\n%s\n", cmp.Diff(appCRNames, expectedAppCRNames))
+			}
+
+			return nil
+		}
+		b := backoff.NewMaxRetries(150, 5*time.Second)
```
TheoBrigitte: Why do we need such a high number of retries?

kopiczko: I wanted to have 150s in here. 2m + some extra for resync period.

TheoBrigitte: Ok this sounds fair, but what I understood from this code is that there will be up to 150 retry, each with an interval of 5s. 150*5=750s ~= 12m

PR #111, opened 2019-05-07
multireleasehandling: implement second part of the test

kopiczko:


integration/test/multireleasehandling/multi_release_handling_test.go
```diff
@@ -70,13 +70,32 @@ var releaseCR2 = &releasev1alpha1.Release{
...
+	},
+	Spec: releasev1alpha1.ReleaseCycleSpec{
+		DisabledDate: releasev1alpha1.DeepCopyDate{Time: time.Date(2019, 4, 8, 0, 0, 0, 0, time.UTC)},
+		EnabledDate:  releasev1alpha1.DeepCopyDate{Time: time.Date(2019, 1, 8, 0, 0, 0, 0, time.UTC)},
+		Phase:        releasev1alpha1.CyclePhaseEOL,
+	},
+}
+
+// TestMultiReleaseHandling makes sure that App CRs for components shared
+// between EOL and non-EOL Release CRs are not removed.
```
kopiczko: @xh3b4sd let me know if this is what you're looking for.

xh3b4sd: > TestMultiReleaseHandling makes sure that App CRs for components shared between different Releases are not removed when one release transitions to its final EOL phase.

What about this? The EOL/non-EOL lingo is something you need to know. When I look at this in half a year it becomes super hard to recall all the details.

kopiczko: Sure. I don't see how one can not be into EOL lingo while trying understand this code though.

kopiczko:
Tests were green I'm running the again to check if timeouts are fine.

integration/test/multireleasehandling/multi_release_handling_test.go
```diff
@@ -170,7 +193,75 @@ func TestMultiReleaseHandling(t *testing.T) {
...
+				"aws-operator.4.7.0",
+				"cert-operator.0.1.0",
+				"release-aws.v6.1.0",
+			}
+
+			sort.Strings(appCRNames)
+			sort.Strings(expectedAppCRNames)
+
+			if !cmp.Equal(appCRNames, expectedAppCRNames) {
+				return microerror.Maskf(waitError, "\n\n%s\nappCRNames = %#v\nexpectedAppCRNames = %#v\n\n", cmp.Diff(appCRNames, expectedAppCRNames), appCRNames, expectedAppCRNames)
```
kopiczko: The cmp.Diff output is not sufficient for me when it comes to slices:

```
{[]string}[?->0]:
        -: <non-existent>
        +: "aws-operator.4.6.0"
```

I extended it so it prints something like:

```
{[]string}[?->0]:
        -: <non-existent>
        +: "aws-operator.4.6.

appCRNames = []string{"aws-operator.4.7.0", "cert-operator.0.1.0"}
expectedAppCRNames = []string{"aws-operator.4.6.0", "aws-operator.4.7.0", "cert-operator.0.1.0"}
```

I'm thinking about creating our own `cmp` library that would use `github.com/google/go-cmp/cmp` under the hood but with more verbose output.

Something like `cmp.Diff(appCRNames, expectedAppCRNames)`. Then it would output:

```
{[]string}[?->0]:
        -: <non-existent>
        +: "aws-operator.4.6.

left = []string{"aws-operator.4.7.0", "cert-operator.0.1.0"}
right = []string{"aws-operator.4.6.0", "aws-operator.4.7.0", "cert-operator.0.1.0"}
```

TheoBrigitte: diff look ok to me, I don't see a need to improves this

kopiczko: Lucky you. I'm freaking out when I see this.

integration/test/multireleasehandling/multi_release_handling_test.go
```diff
@@ -170,7 +192,75 @@ func TestMultiReleaseHandling(t *testing.T) {
...
+			}
+
+			return nil
+		}
+		b := backoff.NewMaxRetries(65, 6*time.Second)
```
kopiczko: I increased that so it's enough time to wait for a second loop to complete on the CI machine. ResyncPeriod is 2m.

PR #112, opened 2019-05-07
e2e: mark the release as enabled and verfiy it

TheoBrigitte:
Towards: giantswarm/giantswarm#5278

Mark the previously created release as enabled by creating a release cycle with phase=enabled.
Verify that the release status is enabled and was reconciled with the values from the release cycle.


integration/test/releasehandling/release_handling_test.go
```diff
@@ -146,4 +160,52 @@ func TestReleaseHandling(t *testing.T) {
...
+
+	// Verify that release was reconciled, status and label should be updated with values from release cycle.
+	{
+		o := func() error {
+			obj, err := config.K8sClients.G8sClient().ReleaseV1alpha1().Releases().Get(releaseCR.Name, metav1.GetOptions{})
+			if err != nil {
+				return microerror.Mask(err)
+			}
+
+			if !obj.Status.Cycle.DisabledDate.Equal(time.Date(2019, 1, 12, 0, 0, 0, 0, time.UTC)) {
```
kopiczko: You can compare obj.Status.Cycle and releaseCycleCR.Spec here. They have the same type.

integration/test/releasehandling/release_handling_test.go
```diff
@@ -146,4 +160,52 @@ func TestReleaseHandling(t *testing.T) {
...
+
+			k := "release-operator.giantswarm.io/release-cycle-phase"
+			v := obj.Labels[k]
+			if v != releasev1alpha1.CyclePhaseEnabled.String() {
+				return microerror.Maskf(waitError, "obj.Labels[%q] = %q, want %q", obj.Labels[k], releasev1alpha1.CyclePhaseEnabled.String())
+			}
+
+			return nil
+		}
+		b := backoff.NewMaxRetries(150, 1*time.Second)
```
kopiczko: I would suggest decreasing the minikube pressure.

```suggestion
		b := backoff.NewMaxRetries(35, 6*time.Second)
```

PR #113, opened 2019-05-07
e2e: releasecycle

TheoBrigitte:

Test that releases App CR is correctly created once a release cycle CR is created.

### DIFF
--- a/integration/test/releasecycle/releasecycle_test.go
+++ b/integration/test/releasecycle/releasecycle_test.go
@@ -72,20 +72,16 @@ func TestReleaseAppCRCreate(t *testing.T) {
 			}
 
 			if !reflect.DeepEqual(appCR.GetName(), expectedAppCR.GetName()) {
-				t.Fatalf("obj.GetName() = %#v, want %#v", appCR.GetName(), expectedAppCR.GetName())
-				return microerror.Mask(waitError)
+				return microerror.Maskf(waitError, "obj.GetName() = %#v, want %#v", appCR.GetName(), expectedAppCR.GetName())
 			}
 			if !reflect.DeepEqual(appCR.GetNamespace(), expectedAppCR.GetNamespace()) {
-				t.Fatalf("obj.GetNamespace() = %#v, want %#v", appCR.GetNamespace(), expectedAppCR.GetNamespace())
-				return microerror.Mask(waitError)
+				return microerror.Maskf(waitError, "obj.GetNamespace() = %#v, want %#v", appCR.GetNamespace(), expectedAppCR.GetNamespace())
 			}
 			if !reflect.DeepEqual(appCR.GetLabels(), expectedAppCR.GetLabels()) {
-				t.Fatalf(">>> obj.GetLabels()\n%#v\n>>> want\n%#v\n", appCR.GetLabels(), expectedAppCR.GetLabels())
-				return microerror.Mask(waitError)
+				return microerror.Maskf(waitError, ">>> obj.GetLabels()\n%#v\n>>> want\n%#v\n", appCR.GetLabels(), expectedAppCR.GetLabels())
 			}
 			if !reflect.DeepEqual(appCR.Spec, expectedAppCR.Spec) {
-				t.Fatalf(">>> obj.Spec\n%#v\n>>> want\n%#v\n", appCR.Spec, expectedAppCR.Spec)
-				return microerror.Mask(waitError)
+				return microerror.Maskf(waitError, ">>> obj.Spec\n%#v\n>>> want\n%#v\n", appCR.Spec, expectedAppCR.Spec)
 			}
 
 			return nil
--- a/integration/test/releasecycle/releasecycle_test.go
+++ b/integration/test/releasecycle/releasecycle_test.go
@@ -3,19 +3,50 @@
 package releasecycle
 
 import (
-	"context"
-	"fmt"
 	"reflect"
 	"testing"
+	"time"
 
 	applicationv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
 	releasev1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/release/v1alpha1"
 	"github.com/giantswarm/backoff"
+	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/release-operator/integration/env"
 	"github.com/giantswarm/release-operator/service/controller/key"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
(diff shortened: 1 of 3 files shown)


integration/test/releasecycle/releasecycle_test.go
```diff
@@ -3,8 +3,94 @@
...
+				return microerror.Maskf(waitError, "obj.GetName() = %#v, want %#v", appCR.GetName(), expectedAppCR.GetName())
+			}
+			if !reflect.DeepEqual(appCR.GetNamespace(), expectedAppCR.GetNamespace()) {
+				return microerror.Maskf(waitError, "obj.GetNamespace() = %#v, want %#v", appCR.GetNamespace(), expectedAppCR.GetNamespace())
+			}
+			if !reflect.DeepEqual(appCR.GetLabels(), expectedAppCR.GetLabels()) {
+				return microerror.Maskf(waitError, ">>> obj.GetLabels()\n%#v\n>>> want\n%#v\n", appCR.GetLabels(), expectedAppCR.GetLabels())
+			}
+			if !reflect.DeepEqual(appCR.Spec, expectedAppCR.Spec) {
+				return microerror.Maskf(waitError, ">>> obj.Spec\n%#v\n>>> want\n%#v\n", appCR.Spec, expectedAppCR.Spec)
```
xh3b4sd: That looks like you want some debug logging in error cases. That should be different to the error though. Please do not mix errors with logging output format.

TheoBrigitte: good point. I splitted up errors and logs. PTALA

integration/test/releasecycle/releasecycle_test.go
```diff
@@ -3,8 +3,98 @@
...
+			if !reflect.DeepEqual(appCR.GetLabels(), expectedAppCR.GetLabels()) {
+				t.Fatalf(">>> obj.GetLabels()\n%#v\n>>> want\n%#v\n", appCR.GetLabels(), expectedAppCR.GetLabels())
+				return microerror.Mask(waitError)
+			}
+			if !reflect.DeepEqual(appCR.Spec, expectedAppCR.Spec) {
+				t.Fatalf(">>> obj.Spec\n%#v\n>>> want\n%#v\n", appCR.Spec, expectedAppCR.Spec)
+				return microerror.Mask(waitError)
+			}
+
+			return err
```
xh3b4sd: This should return nil.

integration/test/releasecycle/releasecycle_test.go
```diff
@@ -3,8 +3,98 @@
...
+			if !reflect.DeepEqual(appCR.GetNamespace(), expectedAppCR.GetNamespace()) {
+				t.Fatalf("obj.GetNamespace() = %#v, want %#v", appCR.GetNamespace(), expectedAppCR.GetNamespace())
+				return microerror.Mask(waitError)
+			}
+			if !reflect.DeepEqual(appCR.GetLabels(), expectedAppCR.GetLabels()) {
+				t.Fatalf(">>> obj.GetLabels()\n%#v\n>>> want\n%#v\n", appCR.GetLabels(), expectedAppCR.GetLabels())
+				return microerror.Mask(waitError)
+			}
+			if !reflect.DeepEqual(appCR.Spec, expectedAppCR.Spec) {
+				t.Fatalf(">>> obj.Spec\n%#v\n>>> want\n%#v\n", appCR.Spec, expectedAppCR.Spec)
```
xh3b4sd: Why should we panic here when this is the callback for the backoff? Aren't these errors intended to occur for some time until there is no issue anymore with the resources?

TheoBrigitte: yeah right, I should use `t.Logf` here in order no to make the test fail. But this would potentially yield many logs lines stating there's an error while this might not be the case, that would be ugly and misleading. Can we just stick with the previous error annotation ? which does exactly what we want: show the error when the backoff fails.

integration/test/releasecycle/releasecycle_test.go
```diff
@@ -3,8 +3,98 @@
...
+			}
+
+			if !reflect.DeepEqual(appCR.GetName(), expectedAppCR.GetName()) {
+				t.Fatalf("obj.GetName() = %#v, want %#v", appCR.GetName(), expectedAppCR.GetName())
+				return microerror.Mask(waitError)
+			}
+			if !reflect.DeepEqual(appCR.GetNamespace(), expectedAppCR.GetNamespace()) {
+				t.Fatalf("obj.GetNamespace() = %#v, want %#v", appCR.GetNamespace(), expectedAppCR.GetNamespace())
+				return microerror.Mask(waitError)
+			}
```
kopiczko: Checking the name and the namespace makes little sense here IMO. The object is "gotten" with those.

integration/test/releasecycle/releasecycle_test.go
```diff
@@ -3,8 +3,98 @@
...
+			if !reflect.DeepEqual(appCR.GetLabels(), expectedAppCR.GetLabels()) {
+				t.Fatalf(">>> obj.GetLabels()\n%#v\n>>> want\n%#v\n", appCR.GetLabels(), expectedAppCR.GetLabels())
+				return microerror.Mask(waitError)
+			}
+			if !reflect.DeepEqual(appCR.Spec, expectedAppCR.Spec) {
+				t.Fatalf(">>> obj.Spec\n%#v\n>>> want\n%#v\n", appCR.Spec, expectedAppCR.Spec)
+				return microerror.Mask(waitError)
+			}
```
kopiczko: I wouldn't focus on the object's content in this case. We can unit-test desired state for that. I would focus on the object existence only.

integration/test/releasecycle/releasecycle_test.go
xh3b4sd: Do what you think is right.

integration/test/releasecycle/releasecycle_test.go
TheoBrigitte: removed those checks

integration/test/releasecycle/releasecycle_test.go
```diff
@@ -3,8 +3,88 @@
...
+		_, err := config.K8sClients.G8sClient().ReleaseV1alpha1().ReleaseCycles().Create(releaseCycleCR)
+		if err != nil {
+			t.Fatalf("err == %v, want %v", err, nil)
+		}
+	}
+
+	// Verifies that release App CR was correctly reconciled.
+	{
+		o := func() (err error) {
+			appCR, err := config.K8sClients.G8sClient().ApplicationV1alpha1().Apps(key.Namespace).Get("release-aws.v6.1.0", v1.GetOptions{})
```
kopiczko:
```suggestion
			appCR, err := config.K8sClients.G8sClient().ApplicationV1alpha1().Apps(expectedAppCR.Namespace).Get(expectedAppCR.Name, v1.GetOptions{})
```

PR #118, opened 2019-05-10
Adds examples of CRs

JosephSalisbury:
Useful to have something in the README when we need it :D


JosephSalisbury:
@giantswarm/team-magic plzzz

PR #120, opened 2019-05-15
Tarball creation in circle-ci

ferrandinand:
Creates a tarball local package

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,7 @@ version: 2.1
 
 orbs:
   app-catalog:
+    working_directory: ~/project
     executors:
       my-executor:
         docker:
@@ -40,6 +41,8 @@ orbs:
         steps:
           - run:
               name: "Creates a tarball file"
+              attach_workspace:
+                at: ~/project
               command: |
                 set -o errexit
                 set -o nounset
@@ -58,6 +61,8 @@ orbs:
                 readonly helm_dir=helm/${CIRCLE_PROJECT_REPONAME}-chart
                 readonly helm_url=https://storage.googleapis.com/kubernetes-helm
                 readonly helm_tarball=helm-v2.11.0-linux-amd64.tar.gz
+
+                readonly last_tag=$(git describe --tags $(git rev-list --tags --max-count=1))
       
                 setup_helm_client() {
                   echo "installing helm binary"
@@ -70,18 +75,34 @@ orbs:
 
                 setup_helm_client
 
-                sed -i -e "s/\[\[ \.VERSION \]\]/${CIRCLE_SHA1}/g" $(find $helm_dir -type f)
+                if [ $last_tag ];then
+                    version=${last_tag//v}"-"${CIRCLE_SHA1}
+                else
+                    version="0.0.0-"${CIRCLE_SHA1}
+                fi
+
+                echo ${version}
+                sed -i -e "s/\[\[ \.VERSION \]\]/${version}/g" $(find $helm_dir -type f)
 
                 echo $(grep -nr $CIRCLE_SHA1 $helm_dir)
-                helm package ${helm_dir}
+                helm package ${helm_dir} -d ~/project
 
-                echo_green "ok: created chart tarball"        
+                echo_green "ok: created chart tarball"  
+      use-tarball:
+        steps:
+          - run:
+              name: "Uses a tarball file"
+              attach_workspace:
+                at: ~/project
+              command: |
+                ls -la  ~/project/
     jobs:
       checkout:
         executor: my-executor
         steps:
           - checkout
           - create-tarball
+          - use-tarball
 
 
 jobs:
@@ -138,7 +159,7 @@ workflows:
   build_e2e:
     jobs:
       - app-catalog/checkout
-          
+
       - build
 
       - e2eTestBasic:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,11 +44,7 @@ orbs:
                 set -o errexit
                 set -o nounset
                 set -o pipefail
-      
-                readonly HELM_DIR=helm/${CIRCLE_PROJECT_REPONAME}-chart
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -26,7 +26,64 @@ e2eTest: &e2eTest
...
+                  curl --user-agent curl-ci-sync -sSL -o "${HELM_TARBALL}" "${HELM_URL}/${HELM_TARBALL}"
+                  tar xzfv "${HELM_TARBALL}"
+                  PATH="$(pwd)/linux-amd64/:$PATH"
+                  helm init --client-only
+                }
+
+                echo "Setting up Helm package..."
+                setup_helm_client
+
+                sed -i -e "s/\[\[ \.SHA \]\]/${CIRCLE_SHA1}/g" $(find $HELM_DIR -type f)
```
ferrandinand: @kopiczko How do you feel about this?

.circleci/config.yml
```diff
@@ -26,7 +26,64 @@ e2eTest: &e2eTest
...
+          - run:
+              name: "Creates a tarball file"
+              command: |
+                set -o errexit
+                set -o nounset
+                set -o pipefail
+      
+                readonly HELM_DIR=helm/${CIRCLE_PROJECT_REPONAME}-chart
+                readonly HELM_URL=https://storage.googleapis.com/kubernetes-helm
+                readonly HELM_TARBALL=helm-v2.11.0-linux-amd64.tar.gz
```
kopiczko: I'd prefer to:

* Have variables defined below `echo_*` functions as this part is just copy/paste and I'd like to skip that while reading scripts.
* Have scripts variables in lower case. This is to distinguish them from environment variables. This is bash good practice.

.circleci/config.yml
```diff
@@ -26,7 +26,64 @@ e2eTest: &e2eTest
...
+      
+                echo_green() {
+                  echo -e "\e[32m${1}"
```
kopiczko: I updated those:

```suggestion
                  echo -e "\e[32m${1}\e[0m"
```

.circleci/config.yml
```diff
@@ -26,7 +26,64 @@ e2eTest: &e2eTest
...
+                }
+                echo_red() {
+                  echo -e "\e[31m${1}"
```
kopiczko:
```suggestion
                  echo -e "\e[31m${1}\e[0m"
```

.circleci/config.yml
```diff
@@ -26,7 +26,64 @@ e2eTest: &e2eTest
...
+                }
+                echo_yellow() {
+                  echo -e "\e[33m${1}"
```
kopiczko:
```suggestion
                  echo -e "\e[33m${1}\e[0m"
```

.circleci/config.yml
```diff
@@ -26,7 +26,64 @@ e2eTest: &e2eTest
...
+
+                setup_helm_client() {
+                  echo "Setting up Helm client..."
+                  curl --user-agent curl-ci-sync -sSL -o "${HELM_TARBALL}" "${HELM_URL}/${HELM_TARBALL}"
+                  tar xzfv "${HELM_TARBALL}"
+                  PATH="$(pwd)/linux-amd64/:$PATH"
+                  helm init --client-only
+                }
+
+                echo "Setting up Helm package..."
```
kopiczko:
```suggestion
```

.circleci/config.yml
```diff
@@ -26,7 +26,64 @@ e2eTest: &e2eTest
...
+                }
+                echo_red() {
+                  echo -e "\e[31m${1}"
+                }
+                echo_yellow() {
+                  echo -e "\e[33m${1}"
+                }
+
+                setup_helm_client() {
+                  echo "Setting up Helm client..."
```
kopiczko:
```suggestion
                  echo "installing helm binary"
```

.circleci/config.yml
```diff
@@ -26,7 +26,64 @@ e2eTest: &e2eTest
...
+                  curl --user-agent curl-ci-sync -sSL -o "${HELM_TARBALL}" "${HELM_URL}/${HELM_TARBALL}"
+                  tar xzfv "${HELM_TARBALL}"
+                  PATH="$(pwd)/linux-amd64/:$PATH"
+                  helm init --client-only
```
kopiczko:
```suggestion
                  helm init --client-only
                  echo "installed helm binary"
```

.circleci/config.yml
```diff
@@ -26,7 +26,64 @@ e2eTest: &e2eTest
...
+
+                echo "Setting up Helm package..."
+                setup_helm_client
+
+                sed -i -e "s/\[\[ \.SHA \]\]/${CIRCLE_SHA1}/g" $(find $HELM_DIR -type f)
+
+                echo $(grep -nr $CIRCLE_SHA1 $HELM_DIR)
+                helm package ${HELM_DIR}
+
+                echo_green "ok: Created helm package"
```
kopiczko:
```suggestion
                echo_green "ok: created chart tarball"        
```

.circleci/config.yml
kopiczko: I don't have an issue with the command itself, but we should replace `.SHA` with `.VERSION`.

ferrandinand: Using version makes me feel that we will need to change all charts and we can break some backward compability.
Changing SHA for version is just a matter of naming? Or are we replacing the current system x.x.x-SHA for only VERSION which contains only the SHA? With the second one we could lose the possibility of notifying major/minor version. Is this right?

kopiczko: We need to replace current `v1.0.0-[[ .SHA ]]` with `[[ .VERSION ]]`. Otherwise how we can set (git) tagged version?

ferrandinand: I understand now. Makes sense if we are doing that tagging strategy

.circleci/config.yml
ferrandinand: I feel like they are constants more than scripts variables so it would make sense to keep them capitalized if so.

PR #127, opened 2019-07-31
Import app resource from apiextensions 

tomahawk28:
Following changes from https://github.com/giantswarm/apiextensions/pull/265

Reroute `app` resource usage to apiextensions.


pkg/testfixture/app_cr.go
```diff
@@ -34,12 +34,12 @@ func NewAppCRFromFilled(modifyFunc func(*applicationv1alpha1.App)) *applicationv
 			},
 			Name:      "test-spec-name",
 			Namespace: "test-spec-namespace",
-			UserConfig: applicationv1alpha1.AppSpecConfig{
-				ConfigMap: applicationv1alpha1.AppSpecConfigConfigMap{
+			UserConfig: applicationv1alpha1.AppSpecUserConfig{
```
tomahawk28: @oponder Hi, I found out this part giving a build error after vendor update. Just let you know that I fixed it like below.

PR #128, opened 2019-09-04
Add network policy 

glitchcrab:
towards giantswarm/giantswarm/issues/4383

### DIFF
--- a/helm/release-operator-chart/values.yaml
+++ b/helm/release-operator-chart/values.yaml
@@ -17,6 +17,8 @@ resource:
     name: "release-operator-network-policy"
   psp:
     name: "release-operator-psp"
+  pullSecret:
+    name: "release-operator-pull-secret"
   service:
     port: 8000
-    protocol: TCP
+    protocol: "TCP"
--- a/helm/release-operator-chart/values.yaml
+++ b/helm/release-operator-chart/values.yaml
@@ -1,3 +1,8 @@
+name: release-operator
 namespace: giantswarm
+
 userID: 1000
 groupID: 1000
+
+port: 8000
+protocol: TCP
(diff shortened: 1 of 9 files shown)


helm/release-operator-chart/values.yaml
```diff
@@ -1,3 +1,8 @@
+name: release-operator
```
kopiczko: Why we don't do similarly to https://github.com/giantswarm/aws-operator/blob/dbb7c297657d782561f3dd83e567303aace22278/helm/aws-operator/values.yaml#L31-L39 ?

I.e. 

```
resource:
  default:
    name: "release-operator"
    namespace: "giantswarm"
  networkPolicy:
    name: "release-operator-network-policy"
    namespace: "giantswarm"
```

glitchcrab: I can do, i'm just trying not to scope my PRs too wide. will all operators eventually undergo VOO? Or do you mean just to keep things more consistent?

kopiczko: This is the only one that won't so in this case it's consistency only. I'm not to fussed about other project because we will change them anyway during VOO flattening.

helm/release-operator-chart/values.yaml
```diff
@@ -1,3 +1,24 @@
...
+    namespace: "giantswarm"
+  networkPolicy:
+    name: "release-operator-network-policy"
+  psp:
+    name: "release-operator-psp"
+  pullSecret:
+    name: "release-operator-pull-secret"
+  service:
+    port: 8000
+    protocol: "TCP"
```
kopiczko: I didn't mean you need to change everything now, but I appreciate that :)

glitchcrab: yeah i went a bit overboard :grimacing: for any other operators i'll make a followup PR

kopiczko: No need to bother with that. As I said we'll get there when we will be flattening there.

PR #130, opened 2019-09-13
pkg, service, vendor: use appresource from giantswarm/resource

kopiczko:


pkg/testfixture/app_cr.go
```diff
@@ -16,7 +16,7 @@ func NewAppCRFromFilled(modifyFunc func(*applicationv1alpha1.App)) *applicationv
 		},
 		Spec: applicationv1alpha1.AppSpec{
 			Catalog: "test-spec-catalog",
-			Config: applicationv1alpha1.AppSpecConfig{
+			Config: &applicationv1alpha1.AppSpecConfig{
```
rossf7: These pointers still worry me. As I think it can cause bugs due to the nil handling. But no blocker I'd just like to discuss again next week.

kopiczko: Agreed. I just want to push the vendor.

PR #135, opened 2019-11-21
Vendor dependencies, update k8s to 1.16.3

sslavic:
Towards giantswarm/giantswarm#7675


rossf7:
Tested on `geckon`.

One e2e test is red but operator is not in production usage yet. Going with this.

PR #136, opened 2019-12-06
Use k8sclient collection

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/7963.


xh3b4sd:
@kopiczko where and how can I test `release-operator`? On `gauss` there are no releases.

kopiczko:
I see E2E are green so it's fine to merge (`e2eTestReleaseHandling` is flapping).

xh3b4sd:
Do you have it on your list to fix the flapping? I didn't see something obviously wrong when I checked the code but this situation is really unfortunate. I was spending quite some time here.

kopiczko:
Nothing to worry here. We'll work on it when we get back to this operator.

kopiczko:
But this is weird that this doesn't work. Because the test itself seems to look fine.

PR #140, opened 2020-03-02
Update apiextensions & remove old implementation

MarcelMue:


MarcelMue:
cc @giantswarm/team-magic Removing this stuff here as it's no longer matching the release CR we have. Revert if you think this is wrong.

kopiczko:
Are you going to implement something different here?

MarcelMue:
> Are you going to implement something different here?

Not currently - we mainly needed the operator to not destroy anything when the new CRD is on CPs.

kopiczko:
I see. 👍

PR #149, opened 2020-04-02
set dev version 0.2.3-dev

xh3b4sd:


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA      = "n/a"
 	name        = "release-operator"
 	source      = "https://github.com/giantswarm/release-operator"
-	version     = "0.2.1"
+	version     = "0.2.1-dev"
```
MarcelMue, line 8:
```suggestion
	version     = "0.2.2-dev"
```

Version number should increase - 0.2.1 is released already.

PR #175, opened 2020-06-30
Remove duplication of apps

yulianedyalkova:
Do not attempt to create the same app twice if it is referenced more than once in the releases

### DIFF
--- a/service/controller/key/key.go
+++ b/service/controller/key/key.go
@@ -36,11 +36,28 @@ func BuildAppName(operator releasev1alpha1.ReleaseSpecComponent) string {
 func ExtractOperators(comps []releasev1alpha1.ReleaseSpecComponent) []releasev1alpha1.ReleaseSpecComponent {
 	var operators []releasev1alpha1.ReleaseSpecComponent
 	for _, c := range comps {
-		if strings.Contains(c.Name, "operator") && c.Name != "chart-operator" && c.Name != "app-operator" {
+		if IsOperator(c) {
 			operators = append(operators, c)
 		}
 	}
 	return operators
+}
+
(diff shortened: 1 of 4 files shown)


service/controller/key/key.go
```diff
@@ -43,18 +43,33 @@ func ExtractOperators(comps []releasev1alpha1.ReleaseSpecComponent) []releasev1a
...
+		for _, component := range release.Spec.Components {
+			if IsOperator(component) && (operators[BuildAppName(component)] == releasev1alpha1.ReleaseSpecComponent{}) {
+				operators[BuildAppName(component)] = component
+			}
+		}
+	}
+	return operators
+}
+
+func IsOperator(component releasev1alpha1.ReleaseSpecComponent) bool {
```
MarcelMue, line 59: We should use this function in line 39 in the `ExtractOperators` function.

yulianedyalkova: That's why I added the function, I just forgot to do it.

service/controller/key/key.go
```diff
@@ -36,25 +36,40 @@ func BuildAppName(operator releasev1alpha1.ReleaseSpecComponent) string {
...
 	}
 	return comp.Version
 }
 
-func AppReferenced(operators []releasev1alpha1.ReleaseSpecComponent, app applicationv1alpha1.App) bool {
-	for _, operator := range operators {
-		if BuildAppName(operator) == app.Name && GetOperatorRef(operator) == app.Spec.Version {
-			return true
-		}
+func AppReferenced(operators map[string]releasev1alpha1.ReleaseSpecComponent, app applicationv1alpha1.App) bool {
```
MarcelMue, line 70: In a follow up - let's sort functions in this file alphabetically.

yulianedyalkova: That was a part of my next PR already

PR #178, opened 2020-07-01
Add metrics

stone-z:
## Checklist


helm/release-operator/templates/configmap.yaml
```diff
@@ -12,6 +12,8 @@ data:
       listen:
         address: 'http://0.0.0.0:8000'
     service:
+      installation:
+        name: '{{ .Values.Installation.V1.Name }}'
```
stone-z, line 16: This will be removed again

stone-z:
Metrics are being collected on godsmack

service/collector/release.go
```diff
@@ -0,0 +1,102 @@
...
+		return microerror.Mask(err)
+	}
+
+	for _, release := range releases.Items {
+		ch <- prometheus.MustNewConstMetric(
+			ReleaseDesc,
+			prometheus.GaugeValue,
+			gaugeValue,
+			release.Name,
+			release.Namespace,
```
tfussell, line 95: Releases are cluster-scoped so we don't need to export the namespace (I think it would just be the empty string). See https://github.com/giantswarm/apiextensions/blob/master/config/crd/v1/release.giantswarm.io_releases.yaml#L18

service/collector/release.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+const (
+	namespace = "release_operator"
+	subsystem = "release"
+)
+
+var (
+	ReleaseDesc *prometheus.Desc = prometheus.NewDesc(
+		// TODO: What is the most descriptive name for this? Maybe change to "states" or "statuses"?
```
tfussell, line 25: How about naming it "releases" since it tells the name, the state (from spec), and the readiness (from status) of all releases? I can imagine adding metrics about other resources to the operator in the future.

stone-z: The current metric name when concatenated is "release_operator_release_info", so changing "info" to "releases" would yield "release_operator_release_releases" XD 

I think the fact that it's about releases is covered. "Info" is what most other operators are doing, but just seems very unspecific to me, so I was leaning "states". Then the full metric name would be "release_operator_release_states". Open to other thoughts

tfussell: I see what you mean. The only important part of the metric in my opinion is whether the release is satisfied or not, so I think `release_operator_release_status` would be a reasonable name.

service/collector/release.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func (r *ReleaseCollector) collectReleaseStatus(ctx context.Context, ch chan<- prometheus.Metric) error {
+	releases, err := r.k8sClient.G8sClient().ReleaseV1alpha1().Releases().List(metav1.ListOptions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	for _, release := range releases.Items {
+		ch <- prometheus.MustNewConstMetric(
```
yulianedyalkova, line 90: Is there a reason that we're using `MustNewConstMetric` instead of `NewConstMetric`? Does it make sense to log and propagate the error back instead of directly panicking here?

service/collector/release.go
stone-z: Yeah, I was on the fence anticipating a hypothetical case (e.g. for prow), but it's removed now.

service/collector/release.go
stone-z: That sounds good, going with that

service/collector/release.go
stone-z: Honestly not sure, it seems to be like this everywhere else we collect metrics this way but I don't know if that's intentional. It probably would be better to log and recover but let's ask @giantswarm/sig-monitoring if there's a reason to do it this way.

JosephSalisbury: hi hi hi

from https://godoc.org/github.com/prometheus/client_golang/prometheus#NewConstMetric:
```
NewConstMetric returns an error if the length of labelValues is not consistent with the variable labels in Desc or if Desc is invalid
```

On the whole, that's pretty hard to fuck up. Sure, I don't think we have any 100% rule that you must use `MustNewConstMetric` over `NewConstMetric`, but I find it only really errors if you've just written the thing wrong, as opposed to some runtime issue.

yulianedyalkova: In this way we kind of depend on that behaviour not changing, but it's probably fine in this case. Especially if we're using it everywhere already 🤷‍♀️ I was just curious

stone-z:
Going with this if there is no more feedback

PR #180, opened 2020-07-02
Check catalog when comparing apps

yulianedyalkova:

### DIFF
--- a/service/controller/key/key.go
+++ b/service/controller/key/key.go
@@ -26,7 +26,8 @@ const (
 )
 
 func AppReferenced(app applicationv1alpha1.App, operators map[string]releasev1alpha1.ReleaseSpecComponent) bool {
-	if operator, ok := operators[app.Name]; ok && IsSameApp(operator, app) {
+	operator, ok := operators[app.Name]
+	if ok && IsSameApp(operator, app) {
 		return true
 	}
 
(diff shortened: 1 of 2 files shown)


service/controller/key/key.go
```diff
@@ -94,9 +94,16 @@ func IsOperator(component releasev1alpha1.ReleaseSpecComponent) bool {
 	return strings.HasSuffix(component.Name, "-operator") && component.Name != "chart-operator" && component.Name != "app-operator"
 }
 
+func IsSameApp(component releasev1alpha1.ReleaseSpecComponent, app applicationv1alpha1.App) bool {
+	return BuildAppName(component) == app.Name &&
+		component.Catalog == app.Spec.Catalog &&
+		GetOperatorRef(component) == app.Spec.Version
+
```
MarcelMue, line 101:
```suggestion

```

service/controller/key/key.go
```diff
@@ -26,7 +26,7 @@ const (
 )
 
 func AppReferenced(app applicationv1alpha1.App, operators map[string]releasev1alpha1.ReleaseSpecComponent) bool {
-	if operator, ok := operators[app.Name]; ok && GetOperatorRef(operator) == app.Spec.Version {
+	if operator, ok := operators[app.Name]; ok && IsSameApp(operator, app) {
```
MarcelMue, line 29: I would prefer to split this into 2 expressions:
```
operator, ok := operators[app.Name]
If ok && IsSameApp(operator, app){
...
}
Might just be personal preference though.

PR #181, opened 2020-07-03
Stop using the term operator for the release components

yulianedyalkova:


service/controller/key/key.go
```diff
@@ -50,41 +50,41 @@ func ConstructApp(operator releasev1alpha1.ReleaseSpecComponent) applicationv1al
...
 			Namespace: Namespace,
-			Version:   GetOperatorRef(operator),
+			Version:   GetComponentRef(component),
 		},
 	}
 }
 
-func ExtractAllOperators(releases releasev1alpha1.ReleaseList) map[string]releasev1alpha1.ReleaseSpecComponent {
-	var operators = make(map[string]releasev1alpha1.ReleaseSpecComponent)
+func ExtractAllRelevantComponents(releases releasev1alpha1.ReleaseList) map[string]releasev1alpha1.ReleaseSpecComponent {
```
MarcelMue, line 64: I am not sure if `RelevantComponents` is much better here - how about `ControlPlaneComponents`?

yulianedyalkova: `ControlPlaneComponents ` would be a bit misleading since we are not extracting all of the CP components but only the ones that we care about (thus relevant).

MarcelMue: I don't have a strong opinion either way. Either name isn't fantastic to me.

yulianedyalkova: Do you hate just `Components` less? I'm trying to get rid of `Operators` as a preparation for the next step where the release-operator won't know that it's extracting operators any more.

MarcelMue: Just components is fine by me.

PR #183, opened 2020-07-20
Add documentation for new flow

yulianedyalkova:
## Checklist


README.md
```diff
@@ -5,36 +5,25 @@
...
-      kind: "changed"
+  apps:
+  - name: app-1
+    version: 1.2.1
   components:
-    - name: "azure-operator"
-      version: "0.0.1"
-  parentVersion: "0.0.0"
-  version: "0.0.1"
+  - name: ignore-me
```
tfussell, line 18: Could add `// releaseOperatorDeploy: false` just to make it clear why it's being ignored.

docs/workflow.md
```diff
@@ -0,0 +1,75 @@
...
+  date: "2020-04-27T12:00:00Z"
+  state: active
+```
+
+The only part of the release that this operator cares for is the list of components in the spec. It will make sure to deploy all of the components that have `releaseOperatorDeploy` set to true. So in the example above, it will ignore `ignore-me` and deploy `deploy-me` on the CP.
+
+#### What does deploying mean?
+
+So as you can see from the example above, each component can have a subset of the following fields:
+* `catalog`: which catalog to take the component from? (e.g. control-plane-catalog, control-plane-test-catalog)
```
tfussell, line 33: This defaults to `control-plane-catalog`, right? Might be good to mention that this will be set if not specified.

docs/workflow.md
```diff
@@ -0,0 +1,75 @@
...
+* `nam`e: name of the component.
```
tfussell, line 34:
```suggestion
* `name`: name of the component.
```

docs/workflow.md
```diff
@@ -0,0 +1,75 @@
...
+* `referenc`e: reference of the component. Enables testing of untagged commits/branches.
```
tfussell, line 35:
```suggestion
* `reference`: A reference can be a tag, commit, or branch. Used for testing and alternative tags for existing versions.
```

docs/workflow.md
```diff
@@ -0,0 +1,75 @@
...
+* `releaseOperatorDeploy`: controls if this operator will deploy the component.
+* `version`: version of the component.
+
+On each reconcile loop release-operator does the following:
```
tfussell, line 39:
```suggestion
On each reconciliation loop, release-operator does the following:
```

docs/workflow.md
```diff
@@ -0,0 +1,75 @@
...
+1. Iterate over all of the releases on the CP.
```
tfussell, line 40:
```suggestion
1. Iterate over all of the Release CRs on the CP.
```

docs/workflow.md
```diff
@@ -0,0 +1,75 @@
...
+1. Undeploy all components currently not referred by any release.
```
tfussell, line 41:
```suggestion
1. Undeploy all components currently not referenced by any release.
```

docs/workflow.md
```diff
@@ -0,0 +1,75 @@
...
+1. For each release, deploy all the components that have `releaseOperatorDeploy` set to `true`.
```
tfussell, line 42:
```suggestion
1. Deploy all components referenced by at least one release that have `releaseOperatorDeploy` set to `true`.
```

docs/workflow.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+A few key points here:
+* the app name is a concatenation of the component name and version.
+* reference is being passed through as version in the App CR. If no reference is being used, then release-operator will default to using the component version.
+* for every app, `inCluster` is being set to `true` in the `kubeConfig`.
+
+#### Release status
+
+In the releases's status, you can find a `Ready` field that will tell you the current state of the release. Once a release has all of its components deployed (even the ones that release-operator is not accountable for), the value of `Ready` will change to `true`.
```
tfussell, line 72:
```suggestion
In the releases's status, you can find a `Ready` field that will tell you the current state of the release. Once a release has all of its components
deployed (even the ones that release-operator is not accountable for), the value of `Ready` will be updated to to `true`.
```

tfussell: Probably worth noting that release-operator only ensures that the App CR exists in the API. App-operator and chart-operator take care of actually deploying the corresponding helm chart.

docs/workflow.md
```diff
@@ -0,0 +1,75 @@
...
+
+The status of a release is also being exported as a Prometheus metric.
```
tfussell, line 74: I think it would be cool to describe when the alert will fire and maybe even link to the ops-recipe. Up to you though.

docs/workflow.md
yulianedyalkova: Yes, but the defaulting happens on the CRD level and I didn't want to couple the doc with that (it could easily become outdated). As far as release-operator cares, it just gets a catalog and passes it through.

docs/workflow.md
yulianedyalkova: I think the reference could only be in the `X.Y.Z` or `X.Y.Z-SHA` format. You can pass a tag/commit/branch to the `opsctl` command and it does the translation.

README.md
yulianedyalkova: Updated.

docs/workflow.md
yulianedyalkova: Updated.

docs/workflow.md
yulianedyalkova: Updated.

docs/workflow.md
yulianedyalkova: Updated.

docs/workflow.md
yulianedyalkova: Updated.

docs/workflow.md
yulianedyalkova: Updated.

docs/workflow.md
yulianedyalkova: That's a really good point!

docs/workflow.md
yulianedyalkova: Updated.

docs/workflow.md
tfussell: Makes sense. I would just like to specify here what the reference can be. How about "A reference points to specific version of a component (e.g. `0.1.0`) with an optional SHA suffix (e.g. `0.1.0-1078ad9d2c15178d1466f79f1a54ebd9c92d9614`) to specify a commit. Used for testing and referring to alternative releases of existing components."

PR #204, opened 2020-09-03
Remove unused deprecated releases

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/12624

Removes the apps for a release if the release is deprecated and no running cluster uses that release version.

Still todo:
- [x] make provider-independent (currently hardcoded AWS)
- [x] remove debug output

## Checklist


service/controller/release/resource/apps/resource.go
```diff
@@ -120,6 +136,155 @@ func (r *Resource) ensureState(ctx context.Context) error {
...
+		return nil, microerror.Mask(err)
+	}
+
+	var clusters []TenantCluster
+	for _, cluster := range awsclusters.Items {
+		c := TenantCluster{
+			ID:              cluster.Name,
+			OperatorVersion: cluster.Labels[apiexlabels.AWSOperatorVersion],
+			ReleaseVersion:  cluster.Labels[apiexlabels.ReleaseVersion],
+			Provider:        "aws", // TODO: Parameterize or detect
```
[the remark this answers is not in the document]
stone-z: I was thinking about ditching the parameter. Aside from API calls, is there a downside to just always getting all the providers? Every control plane has all the resources, the lists will just come back empty. I kind of like the idea of keeping the logic identical across providers

service/controller/release/resource/apps/resource.go
```diff
@@ -120,6 +136,155 @@ func (r *Resource) ensureState(ctx context.Context) error {
...
+	return ""
+}
+
+// Takes a list of tenant clusters and returns two maps containing the versions of their release and operator versions.
+func consolidateClusterVersions(clusters []TenantCluster) (map[string]bool, map[string]bool) {
+	releaseVersions := make(map[string]bool)
+	operatorVersions := make(map[string]bool)
+
+	for _, c := range clusters {
+		fmt.Printf("Cluster %s (%s) is using operator version %s\n", c.ID, c.ReleaseVersion, c.OperatorVersion)
```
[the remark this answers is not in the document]
stone-z: Yeah this'll go away, just for my debugging

service/controller/release/resource/apps/resource.go
```diff
@@ -120,6 +136,155 @@ func (r *Resource) ensureState(ctx context.Context) error {
...
+func (r *Resource) getCurrentTenantClusters() ([]TenantCluster, error) {
+
+	var tenantClusters []TenantCluster
+	{
+		awsClusters, err := r.getCurrentAWSClusters()
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+		tenantClusters = append(tenantClusters, awsClusters...)
+		r.logger.Log("level", "debug", "message", fmt.Sprintf("found %d aws tenant clusters", len(awsClusters)))
```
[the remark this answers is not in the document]
stone-z: Yeah, also just for my sanity checks

go.mod
```diff
@@ -3,21 +3,27 @@ module github.com/giantswarm/release-operator
 go 1.13
 
 require (
-	github.com/giantswarm/apiextensions v0.4.20
+	github.com/giantswarm/apiextensions/v2 v2.1.1-0.20200911125415-95df6338b163
```
stone-z, line 6: Will change once branch is merged

helm/release-operator/templates/rbac.yaml
```diff
@@ -11,6 +11,26 @@ rules:
...
+    resources:
+      - awsclusters
+    verbs:
+      - list
+  - apiGroups:
+      - infrastructure.cluster.x-k8s.io
+    resources:
+      - azureclusters
+    verbs:
+      - "list"
```
MarcelMue, line 25:
```suggestion
      - list
```

service/controller/release/resource/status/cluster.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Returns a list of legacy clusters based on <provider>config resources.
+func (r *Resource) getLegacyClusters(ctx context.Context) ([]TenantCluster, error) {
+	var legacyClusters []TenantCluster
+	aws, err := r.getLegacyAWSClusters(ctx)
+	if IsResourceNotFound(err) {
+		// Fall through
+	} else if err != nil {
+		r.logger.Log("level", "error", "message", fmt.Sprintf("error getting aws legacy clusters: %s", err))
```
MarcelMue, line 107: Why only log the error here?

service/controller/release/resource/status/create.go
```diff
@@ -41,6 +43,40 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		}
 	}
 
+	// Doing this per-release isn't ideal, can we pass this list to each status reconcile somehow?
+	r.logger.LogCtx(ctx, "level", "debug", "message", "searching for running tenant clusters")
```
MarcelMue, line 47: Let's put this log message inside the block - similar to how it is below.

service/service.go
```diff
@@ -81,6 +82,7 @@ func New(config Config) (*Service, error) {
 			SchemeBuilder: k8sclient.SchemeBuilder{
 				appv1alpha1.AddToScheme,
 				releasev1alpha1.AddToScheme,
+				capzv1alpha3.AddToScheme,
```
MarcelMue, line 85: Why is this only necessary for this type?

service/controller/release/resource/status/cluster.go
stone-z: Because IMO we should still try to get the other provider types

MarcelMue: I see, I would have expected any error we want to continue on to go into the fall through. Now we take essentially the same action in both cases of the `if` statement.

Could we then change the log level?

service/service.go
stone-z: `AWSClusters` are already exposed by `G8sClient().Infrastructurev1alpha2()` but `AzureClusters` use the upstream CAPZ type with `CtrlClient()`

service/controller/release/resource/status/cluster.go
stone-z: Change the log level to what? The `IsResourceNotFound` case is expected 2/3 of the time so I don't think we should log those. 

The intention here was to avoid a case where for example the operator tries to get `awsconfig`s on a KVM installation but it fails for some other reason, and then the operator never gets to the `kvmconfigs`. We should not eat that error, but we also probably don't want to give up.

service/controller/release/resource/status/create.go
stone-z: Good idea

service/controller/key/key.go
```diff
@@ -20,10 +20,17 @@ const (
 	LabelAppOperatorVersion = "app-operator.giantswarm.io/version"
 	LabelManagedBy          = "giantswarm.io/managed-by"
 	LabelServiceType        = "giantswarm.io/service-type"
+	LabelKVMOperator        = "kvm-operator.giantswarm.io/version" // TODO: Why isn't this in apiextensions?
```
stone-z, line 23: This has been added to apiextensions -- will be removed from here once apiextensions is released

service/controller/release/resource/status/cluster.go
```diff
@@ -0,0 +1,213 @@
...
+// Takes a list of tenant clusters and returns two maps containing the versions of their release and operator versions.
+func consolidateClusterVersions(clusters []TenantCluster) (map[string]bool, map[string]map[string]bool) {
+	releaseVersions := make(map[string]bool)
+	operatorVersions := make(map[string]map[string]bool)
+
+	// operatorVersions is a nested map including the operator name and version
+	// e.g. operatorVersions["aws-operator"]["8.7.6"]:true
+
+	for _, c := range clusters {
+		fmt.Printf("Cluster %s (%s) is using operator version %s\n", c.ID, c.ReleaseVersion, c.OperatorVersion)
```
stone-z, line 31: Will be removed

service/service.go
MarcelMue: Weird - I wonder if that's intentional.

service/controller/release/resource/status/cluster.go
MarcelMue: I am not too fussed about this either way - so not blocking here.

My concern is that we will not notice a true issue here since any error is just logged. If the error has no impact then it should be a debug log IMO. I think we should go into a direction where we can return the error here and fail reconciliation but alert / collect those errors (through log collection, metrics, whatever).

stone-z: I see -- I think this could also return the error as I don't recall seeing it since working out some bugs. I will change this and re-test.

stone-z:
@giantswarm/team-ludacris PTAL -- other PRs are ready to go when this is approved

service/controller/release/resource/status/cluster.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+		tenantClusters = append(tenantClusters, legacyClusters...)
+	}
+
+	return tenantClusters, nil
+}
+
+// Returns a list of AWS clusters according to the awscluster resource.
+func (r *Resource) getCurrentAWSClusters(ctx context.Context) ([]TenantCluster, error) {
+	awsclusters, err := r.k8sClient.G8sClient().InfrastructureV1alpha2().AWSClusters("default").List(ctx, metav1.ListOptions{})
```
tfussell, line 74: I think Firecracker is working on adding multi-namespace support for AWSClusters. Might be worth checking with them how far out this is.

service/controller/release/resource/status/cluster.go
```diff
@@ -0,0 +1,205 @@
...
+		clusters = append(clusters, c)
+	}
+
+	return clusters, nil
+}
+
+// Returns a list of Azure clusters according to the azurecluster resource.
+func (r *Resource) getCurrentAzureClusters(ctx context.Context) ([]TenantCluster, error) {
+	azureClusters := azurecapi.AzureClusterList{}
+	err := r.k8sClient.CtrlClient().List(ctx, &azureClusters)
```
tfussell, line 98: Why does this use `CtrlClient` but `AWSCluster` uses `G8SClient`?

service/controller/release/resource/status/create.go
```diff
@@ -41,6 +43,41 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 		}
 	}
 
+	// Doing this per-release isn't ideal, can we pass this list to each status reconcile somehow?
```
tfussell, line 46: I think the latest `operatorkit` provides some caching primitives. Is that what you were thinking?

service/controller/release/resource/apps/resource.go
```diff
@@ -122,6 +124,21 @@ func (r *Resource) ensureState(ctx context.Context) error {
 	return nil
 }
 
+func (r *Resource) excludeUnusedDeprecatedReleases(releases releasev1alpha1.ReleaseList) releasev1alpha1.ReleaseList {
```
tfussell, line 127: It would be cleaner to accept and return `[]releasev1alpha1.Release` here since we're just using the `Items` field in the `ReleaseList`.

service/controller/release/resource/status/cluster.go
```diff
@@ -0,0 +1,205 @@
...
+
+type TenantCluster struct {
+	ID               string
+	OperatorVersion  string
+	ProviderOperator string
+	ReleaseVersion   string
+}
+
+// Takes a list of tenant clusters and returns two maps containing the versions of their release and operator versions.
+func consolidateClusterVersions(clusters []TenantCluster) (map[string]bool, map[string]map[string]bool) {
```
tfussell, line 22: I don't really like returning multiple unnamed values like this. I think it would be better to pass the maps in as arguments, give them their own type (`type ReleaseUsageMap map[string]bool`), or use named return values.

service/controller/release/resource/status/cluster.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+		operatorVersions[c.ProviderOperator][c.OperatorVersion] = true
+	}
+
+	return releaseVersions, operatorVersions
+}
+
+// Returns a list of tenant clusters currently running on the installation.
+func (r *Resource) getCurrentTenantClusters(ctx context.Context) ([]TenantCluster, error) {
+
```
tfussell, line 43:
```suggestion
```

service/controller/release/resource/status/cluster.go
```diff
@@ -0,0 +1,205 @@
...
+	var tenantClusters []TenantCluster
+	{
+		awsClusters, err := r.getCurrentAWSClusters(ctx)
```
tfussell, line 46: Since `getCurrentAWSClusters`, `getCurrentAzureClusters`, `getLegacyAWS/KVM/AzureClusters`,  all have the same signature, you could loop over them and append them in sequence to cut down on some code.

service/controller/release/resource/status/cluster.go
stone-z: AWS doesn't use upstream CAPI types, they come from apiextensions. Azure uses the upstream CAPZ types

service/controller/release/resource/apps/resource.go
stone-z: From the caller side I think it is nicer to pass the releases around the same way and operate on the whole API type:
```go
releases = excludeDeletedRelease(releases)
releases = r.excludeUnusedDeprecatedReleases(releases)
```
as opposed to
```go
releases.Items = excludeDeletedRelease(releases.Items)
releases.Items = r.excludeUnusedDeprecatedReleases(releases.Items)
```
But let me know if you feel strongly

service/controller/release/resource/status/cluster.go
stone-z: Sure, good idea. I gave them some names

service/controller/release/resource/status/cluster.go
stone-z: Haha I wasn't sure if that was good Go style, it felt too pythonic -- I like that a lot more

tfussell: You tend to be limited in Go due to typing, but since these all return the same type, I don't see a problem with it. The main goal is clarity/readability and I think that is maintained. In the future we can drop the legacy CRs just by removing elements from the slice which is nice.

service/controller/release/resource/apps/resource.go
tfussell: I was thinking of extracting `Items` from the `ReleaseList` earlier on, but this is fine too.

service/controller/release/resource/status/cluster.go
tfussell: Got it.

service/controller/release/resource/status/cluster.go
stone-z: Turns out the solution for this also addresses your comment about the `CtrlClient` -- this will search across namespaces now and is the same structure as for Azure

MarcelMue (approved):
Fine for me when Thomas is happy. We can iterate here.

service/controller/release/resource/status/create.go
stone-z: I looked around in `operatorkit` but can't tell from the docs and code I saw whether it's what I was thinking. The problem I see is that this code runs once for every release, so we wind up doing 5 API calls per release when chances are the clusters won't change between reconciling this release and the next one. It isn't a ton of waste, just not optimal. It would be nice if we could do this once per reconciling ALL the releases instead of per-resource.

stone-z:
@tfussell sorry for the ping -- is this good to go for you?

tfussell:
Checking.

service/controller/release/resource/status/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+	infrastructurev1alpha2 "github.com/giantswarm/apiextensions/v2/pkg/apis/infrastructure/v1alpha2"
+	apiexlabels "github.com/giantswarm/apiextensions/v2/pkg/label"
+	"github.com/giantswarm/microerror"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	azurecapi "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
+
+	"github.com/giantswarm/release-operator/service/controller/key"
+)
+
+type TenantCluster struct {
```
tfussell, line 15: I think this can be made private.

service/controller/release/resource/status/create.go
tfussell: I don't know too much about it. Not a big deal anyways.

service/controller/release/resource/status/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+		return nil, microerror.Mask(err)
+	}
+
+	var clusters []TenantCluster
+	for _, cluster := range awsClusters.Items {
+		c := TenantCluster{
+			ID:               cluster.Name,
+			OperatorVersion:  cluster.Labels[apiexlabels.AWSOperatorVersion],
+			ReleaseVersion:   cluster.Labels[apiexlabels.ReleaseVersion],
+			ProviderOperator: key.ProviderOperatorAWS,
```
tfussell, line 83:
```suggestion
			ProviderOperator: key.ProviderOperatorAWS,
			ReleaseVersion:   cluster.Labels[apiexlabels.ReleaseVersion],
```

tfussell (approved):
Left a few comments about minor things to fix. LGTM otherwise.

stone-z:
Ok, I think we're all happy enough. Let's iterate

PR #211, opened 2020-09-17
Add some tests for deprecated release logic

stone-z:
## Checklist


stone-z:
Might add some more tests later but no need to make this a huge PR

service/controller/unittest/default_k8sclient.go
```diff
@@ -0,0 +1,82 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+type fakeK8sClient struct {
+	ctrlClient client.Client
+	k8sClient  *fakek8s.Clientset
+}
+
+func FakeK8sClient() k8sclient.Interface {
```
tfussell, line 24: You might be able to use https://github.com/giantswarm/k8sclient/blob/master/pkg/k8sclienttest/empty.go for this. Not sure about the scheme part.

stone-z: Yep looks like it -- thanks. I hadn't seen that used anywhere. The full fake client might be necessary to test the cluster listing but I didn't plan to do that right now

tfussell (approved):

PR #214, opened 2020-10-07
Use Cluster CRs instead of provider cluster types

stone-z:
Toward https://github.com/giantswarm/giantswarm/issues/13651

## Checklist


stone-z:
Tested on `gaia` and `godsmack`

MarcelMue (approved):
LGTM if tested - I am not sure if it will work with Azure as they use CAPI v1alpha3 instead of alpha2? Not certain

stone-z:
I tested on Azure by creating a deprecated cluster, seeing the correct operator get deployed, then deleting the cluster and watching the operator get removed, so at least that much works. I'm new to this CAPI business though, maybe a @giantswarm/team-celestial sanity check would be nice. Do I need to use both (CAPI `v1alpha2` for AWS and `v1alpha3` for Azure)?

fiunchinho:
> I tested on Azure by creating a deprecated cluster, seeing the correct operator get deployed, then deleting the cluster and watching the operator get removed, so at least that much works. I'm new to this CAPI business though, maybe a @giantswarm/team-celestial sanity check would be nice. Do I need to use both (CAPI `v1alpha2` for AWS and `v1alpha3` for Azure)?
> 
> Edit: I think I do -- updating this now

Yes, you do.

stone-z:
Checked again on `godsmack` that the Cluster CRs are actually being read

tfussell (approved):
LGTM.

PR #233, opened 2020-11-03
Mark KVM release as "inUse" while draining

tfussell:
When we upgrade a cluster to a new `kvm-operator` version, the old operator is responsible for draining it (see https://github.com/giantswarm/kvm-operator/blob/master/service/controller/drainer_resource_set.go#L30-L450). release-operator deschedules operators that aren't referenced by an active release or a running cluster. In the case of an upgrade, if this is the last cluster of a deprecated release, the old operator is descheduled and can't drain the old node so it can be replaced. The solutions I can thought of would be to keep the old operator around if it's referenced by any node pod or to let the new operator take care of draining the old pod. This PR implements the first option as it seemed more straightforward than reconciling CRs owned by other operators.

Related discussion https://gigantic.slack.com/archives/CHAHY6VMF/p1604367058111900

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Change `inUse` status of KVM releases to `true` when any pods remain for that release to fix an issue with draining 
+  the last cluster of a deprecated release.
+
 ## [2.1.1] - 2020-10-12
 
 ### Changed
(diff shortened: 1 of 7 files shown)


tfussell:
I deployed this on `geckon` where a pod was stuck termination due to the issue described above. This immediately deployed the missing operator which removed the finalizer and unstuck the upgrade.

MarcelMue (approved):
I am okay with this but we need to somehow document this behaviour. I see our future selves seeing this and wondering "WTF was that for?"

tfussell:
> I am okay with this but we need to somehow document this behaviour. I see our future selves seeing this and wondering "WTF was that for?"

I agree. I added a code comment and there's always `git blame`. What else should I do to document it?

stone-z (approved):
Makes sense to me

tfussell:
Documented here https://github.com/giantswarm/giantswarm/pull/14136

PR #266, opened 2021-02-09
Add config resource

corest:
Towards https://github.com/giantswarm/giantswarm/issues/15500

I didn't change app CR creation logic yet. Will add that with separate PR.
This PR mostly replicates apps resources.

---

For release CR

```
apiVersion: release.giantswarm.io/v1alpha1
kind: Release
metadata:
  annotations:
    giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/x
  finalizers:
  - operatorkit.giantswarm.io/release-operator-release
  name: v12.7.1
spec:
  apps:
  - catalog: default
    name: cert-exporter
    version: 1.2.3
  components:
  - catalog: control-plane-catalog
    name: aws-operator
    releaseOperatorDeploy: true
    version: 9.3.1-fix
  date: "2020-12-24T10:20:34Z"
  state: active
status:
  inUse: false
  ready: false
```

it creates config CR

```
apiVersion: core.giantswarm.io/v1alpha1
kind: Config
metadata:
  creationTimestamp: "2021-02-09T16:48:51Z"
  generation: 1
  labels:
    giantswarm.io/managed-by: release-operator
spec:
  app:
    catalog: control-plane-catalog
    name: aws-operator
    version: 9.3.1-fix
```

## Checklist


service/controller/key/key.go
```diff
@@ -39,10 +40,23 @@ func AppReferenced(app applicationv1alpha1.App, components map[string]releasev1a
...
+
+	return false
+}
+
 func BuildAppName(component releasev1alpha1.ReleaseSpecComponent) string {
 	return fmt.Sprintf("%s-%s", component.Name, component.Version)
 }
 
+func BuildConfigName(component releasev1alpha1.ReleaseSpecComponent) string {
+	return fmt.Sprintf("%s-%s", component.Name, component.Version)
```
MarcelMue, line 57: Should we add `release-operator` to the name somehow? Otherwise I could see issues with opsctl / release-operator interfering. Not sure though.

service/controller/release/resource/configs/resource.go
```diff
@@ -0,0 +1,174 @@
...
+			return microerror.Mask(err)
+		}
+
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("created config %#q in namespace %#q", config.Name, config.Namespace))
+	}
+
+	return nil
+}
+
+func (r *Resource) excludeUnusedDeprecatedReleases(releases releasev1alpha1.ReleaseList) releasev1alpha1.ReleaseList {
```
MarcelMue, line 126: Can we move these functions to `controller/key/key.go` and then use the same functions in both resources? IIRC these are already in the other resource, right?

service/controller/key/key.go
corest: Isn't this the same for apps? Config name matches app name. I don't see where this can interfere

MarcelMue: If we stick to that convention in all other places, then yes this should be fine. `opsctl` creates apps usually with the branch name in the app name. 

But you are probably right and I am overthinking here. Just thought it might be beneficial to be more safe.

corest:
@MarcelMue I've addressed most of the comments

service/controller/release/resource_set.go
```diff
@@ -49,6 +63,7 @@ func NewResourceSet(config ResourceSetConfig) ([]resource.Interface, error) {
 	resources := []resource.Interface{
 		statusResource,
 		appsResource,
+		configsResource,
```
MarcelMue, line 66: Just to do this now already: I think this resource should run before the apps resource.

MarcelMue (approved):
Overall LGTM. Would be nice to have at least one additional review.

stone-z (approved):
Minor suggestions but overall ok with me

service/service.go
```diff
@@ -8,6 +8,7 @@ import (
 	appv1alpha1 "github.com/giantswarm/apiextensions/v2/pkg/apis/application/v1alpha1"
 	infrastructurev1alpha2 "github.com/giantswarm/apiextensions/v2/pkg/apis/infrastructure/v1alpha2"
 	releasev1alpha1 "github.com/giantswarm/apiextensions/v2/pkg/apis/release/v1alpha1"
+	corev1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/core/v1alpha1"
```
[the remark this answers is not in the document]
corest: I don't know about if there is anything preventing us, nut definitely not in this PR

service/controller/key/key.go
```diff
@@ -107,7 +164,23 @@ func IsSameApp(component releasev1alpha1.ReleaseSpecComponent, app applicationv1
...
-func ComponentCreated(component releasev1alpha1.ReleaseSpecComponent, apps []applicationv1alpha1.App) bool {
+func IsSameConfig(component releasev1alpha1.ReleaseSpecComponent, config corev1alpha1.Config) bool {
+	_, managedByReleaseOperator := config.Labels[LabelManagedBy]
+	return component.Name == config.Spec.App.Name &&
+		component.Catalog == config.Spec.App.Catalog &&
+		GetComponentRef(component) == config.Spec.App.Version &&
+		managedByReleaseOperator
+}
+
+func IsSameConfigDeployed(component releasev1alpha1.ReleaseSpecComponent, config corev1alpha1.Config) bool {
```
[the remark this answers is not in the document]
corest: This function not even in use actually. It was just replicated from apps resource. Deleted

kubasobon (approved):

service/controller/key/key.go
```diff
@@ -107,7 +164,15 @@ func IsSameApp(component releasev1alpha1.ReleaseSpecComponent, app applicationv1
 		GetComponentRef(component) == app.Spec.Version
 }
 
-func ComponentCreated(component releasev1alpha1.ReleaseSpecComponent, apps []applicationv1alpha1.App) bool {
+func IsSameConfig(component releasev1alpha1.ReleaseSpecComponent, config corev1alpha1.Config) bool {
+	_, managedByReleaseOperator := config.Labels[LabelManagedBy]
```
kopiczko, line 168: This looks like it's only checking label existence but not the value. What's the point?

PR #268, opened 2021-02-11
Create app on config status ready

corest:
Towards https://github.com/giantswarm/giantswarm/issues/15500

Skip app CR creation of config CR has not proper status (which is set by config-controller)
## Checklist


service/controller/release/resource/apps/resource.go
```diff
@@ -105,6 +122,12 @@ func (r *Resource) ensureState(ctx context.Context) error {
 
 	appsToCreate := calculateMissingApps(components, apps)
 	for _, app := range appsToCreate.Items {
+		if !key.AppConfigDeployed(app, configs) {
+			// Skip this app
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("skipping app %#q as its config is not deployed", app.Name))
+			continue
```
MarcelMue, line 128: I would be interested how much longer it takes like this for apps to be created. I would expect a delay of ~2-5 minutes at maximum which is acceptable to me.

MarcelMue (approved):
Fine with me - I assume this was tested on some test installation?

service/controller/release/resource/apps/resource.go
corest: I've tested on kind cluster only because we don't have config-controller config controller ready (omg that naming). TIL status update is not CR update in sense of what operator reconciles

service/controller/key/key.go
```diff
@@ -31,6 +31,22 @@ const (
 	ProviderOperatorKVM   = "kvm-operator"
 )
 
+func AppConfigDeployed(app applicationv1alpha1.App, configs corev1alpha1.ConfigList) bool {
+	for _, config := range configs.Items {
+		_, configManagedByReleaseOperator := config.Labels[LabelManagedBy]
```
kopiczko, line 36: Again this only check the annotation existence. It would be good compare the value to `project.Name()` IMO.

service/controller/release/resource/apps/resource.go
kopiczko: Creating config takes ~10-15s with cold cache.

kopiczko: > TIL status update is not CR update in sense of what operator reconciles

Yes that's a different resource. I don't know if you can watch that.

service/controller/key/key.go
corest: ah, true

corest:
Tested. Works fine

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `Config` CR resource management into the `release` controller.
 
+### Changed
+
+- Skip `App` CR resource creation if corresponding `Config` CR has no expected status.
+
```
kopiczko, line 19: We don't have to mention that as the support for the Config CR is added within the release.

```suggestion
```

service/controller/key/key.go
```diff
@@ -18,9 +18,10 @@ const (
 	// Namespace is the namespace where App CRs are created.
 	Namespace = "giantswarm"
 
-	LabelAppOperatorVersion = "app-operator.giantswarm.io/version"
-	LabelManagedBy          = "giantswarm.io/managed-by"
-	LabelServiceType        = "giantswarm.io/service-type"
+	LabelAppOperatorVersion    = "app-operator.giantswarm.io/version"
+	LabelConfigOperatorVersion = "config-controller.giantswarm.io/version"
```
kopiczko, line 22: This annotation is available in apiextensions. IMO it should be imported from there.

service/controller/key/key.go
```diff
@@ -158,18 +160,41 @@ func GetProviderOperators() []string {
 	return []string{ProviderOperatorAWS, ProviderOperatorAzure, ProviderOperatorKVM}
 }
 
+func GetAppConfig(app applicationv1alpha1.App, configs corev1alpha1.ConfigList) (
+	corev1alpha1.ConfigStatusConfigConfigMapRef, corev1alpha1.ConfigStatusConfigSecretRef) {
```
kopiczko, line 164: Why not returning `corev1alpha1.ConfigStatusConfigConfig` instead?

service/controller/key/key.go
```diff
@@ -158,18 +160,41 @@ func GetProviderOperators() []string {
...
+
+		if app.Name == config.Name &&
+			app.Spec.Name == config.Status.App.Name &&
+			app.Spec.Version == config.Status.App.Version &&
+			app.Spec.Catalog == config.Status.App.Catalog &&
+			configIsManagedByReleaseOperator &&
+			configManagedByLabel == project.Name() {
+			cmRef = config.Status.Config.ConfigMapRef
+			secretRef = config.Status.Config.SecretRef
+		}
```
kopiczko, line 179:
```suggestion
		matches := true
		matches = matches && app.Name == config.Name
		matches = matches && app.Spec.Name == config.Status.App.Name
		matches = matches && app.Spec.Version == config.Status.App.Version
		matches = matches && app.Spec.Catalog == config.Status.App.Catalog
		matches = matches && configIsManagedByReleaseOperator
		matches = matches && configManagedByLabel == project.Name()

		if matches {
			cmRef = config.Status.Config.ConfigMapRef
			secretRef = config.Status.Config.SecretRef
			break
		}
```

service/controller/release/resource/apps/resource.go
```diff
@@ -105,6 +122,18 @@ func (r *Resource) ensureState(ctx context.Context) error {
 
 	appsToCreate := calculateMissingApps(components, apps)
 	for _, app := range appsToCreate.Items {
+		cmRef, secretRef := key.GetAppConfig(app, configs)
+		if cmRef.Name == "" && secretRef.Name == "" {
```
kopiczko, line 126: That does not happen. App status and CM and Secret ref are set in a single request in config-controller. But I'm ok to keep that logic.

kopiczko (approved):

PR #314, opened 2021-05-26
Fix App CR creation with proper references

corest:
I'll clean up failed app CRs after releasing this

But omg, how did I test this to end up with such a stupid mistake 

## Checklist


whites11 (approved):
LGTMIIWFY

PR #406, opened 2022-01-24
Fix v3 for package

njuettner:
We updated to v3 a long time ago but forgot to update it modules. because we moved release apis from apiextensions now to release-operator we need to fix that.


njuettner:
Dup #406 closing this one

PR #432, opened 2022-03-25
rename vmware to vsphere

glitchcrab:
towards https://github.com/giantswarm/giantswarm/issues/21459

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,16 +104,3 @@ workflows:
               ignore: /.*/
             tags:
               only: /^v.*/
-
-      - architect/push-to-app-collection:
-          context: architect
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -107,9 +107,9 @@ workflows:
 
       - architect/push-to-app-collection:
```
tfussell, line 108: Can we delete this instead? We don't need release-operator for pure CAPI clusters.

glitchcrab: sure, i was just doing an automated sed/gh pr create for every repo which pushes to the vsphere collection. happy to remove instead though.

tfussell (approved):

